PR #2, opened 2025-09-12
Initial SDK Implementation: Dataset Recording and Replay for Trossen AI  Kit

shantanuparab-tr:
This pull request introduces the first complete implementation of the **Trossen Robotics Data Collection SDK**, designed for recording and replaying robot episodes using the **Stationary AI Kit**. The SDK enables structured dataset creation that is fully compatible with the **Hugging Face LeRobot** format, supporting downstream use in training, evaluation, visualization, and sharing workflows.

---

### Dataset Recording Pipeline

* Records joint observations and actions with timestamps.
* Captures episodes as partitioned **Parquet** datasets for performance and scalability.
* Supports multi-camera setups using **asynchronous image writers** to maintain real-time performance.
* Enables parallel video encoding for efficient dataset post-processing.

### Episode Replay

* Replays actions from Parquet files, replicating the timing of original demonstrations.
* Streams joint positions to the robot in real time for accurate policy evaluation.

### LeRobot-Compatible Format

* Dataset structure and metadata conform to LeRobot’s open standard.
* Compatible with Hugging Face tools for visualization, training, evaluation, and upload.

---

## SDK Architecture

* Modular design for future extensions (e.g., Solo and Mobile AI Kits).
* Structured repository layout for managing episodes, tasks, statistics, and metadata.
* Implemented in C++ with emphasis on performance and extensibility.
* Built-in metadata generation for integration with machine learning workflows.

---

## Design Considerations

* Heavily inspired by the **LeRobot platform** from Hugging Face to maximize ecosystem compatibility.
* Aligns with the formats used for ACT, Pi-0, and other models used in robotic learning.
* Intended to support scalable and reproducible dataset pipelines, both locally and for community use.

---

## Notes

* This is a **foundational implementation** intended to be reviewed and extended in future pull requests.
* Folder structure, naming conventions, and CMake configuration will be refined over time.
* Additional features such as sensor fusion, robot configuration flexibility, dataset statistics, and unit tests will be added in follow-up commits.


include/trossen_ai_robot_devices/trossen_ai_cameras.hpp
```diff
@@ -0,0 +1,174 @@
...
+
+/// @brief Data structure to hold image data from the camera including depth map if available in OpenCV Mat format
+struct ImageData {
+    /// @brief Name of the camera that captured the image
+    std::string camera_name;
+    /// @brief OpenCV Mat to hold the image data
+    cv::Mat image;
+    /// @brief OpenCV Mat to hold the depth information
+    cv::Mat depth_map;
+};
```
lukeschmitt-tr: What's the reason for holding both of these? Can we immediately convert the rs2::frame into a cv::Mat in the format we expect with all the relevant data instead?

include/trossen_ai_robot_devices/trossen_ai_cameras.hpp
```diff
@@ -0,0 +1,174 @@
...
+         * @return ColorDepthData structure containing color image and depth map frames
+         */
+        trossen_ai_robot_devices::ColorDepthData read() ;
+
+        /**
+         * @brief Asynchronously read a frame from the camera
+         * @return ImageData structure containing color image and depth map
+         */
+        trossen_ai_robot_devices::ImageData async_read();
+
```
lukeschmitt-tr: It is more efficient (for large variables) to pass a reference to methods like this as it avoids copies. You can also then reserve the return type for boolean returns to see if the operation succeeded instead of checking for special null values or try/catching.

```
bool async_read(trossen_ai_robot_devices::ImageData & data);
```

include/trossen_ai_robot_devices/trossen_ai_cameras.hpp
```diff
@@ -0,0 +1,174 @@
...
+        /// @brief Flag to indicate if depth map should be captured
+        bool use_depth_ {false};
+        /// @brief Realsense camera pipeline
+        rs2::pipeline camera_;
+    };
+
+
+
+
+class TrossenAsyncImageWriter {
```
lukeschmitt-tr: These don't have to be named Trossen*

```suggestion
class AsyncImageWriter {
```

include/trossen_ai_robot_devices/trossen_ai_cameras.hpp
```diff
@@ -0,0 +1,174 @@
...
+     * @brief Push an image to be written asynchronously
+     * @param image OpenCV Mat containing the image data
+     * @param filename Path where the image should be saved
+     * Adds the image and filename to the queue for asynchronous writing
+     */
+    void push(const cv::Mat& image, const std::string& filename);
+
+private:
+    /// @brief Queue to hold images and their filenames
+    std::queue<std::pair<cv::Mat, std::string>> image_queue_;
```
lukeschmitt-tr: Creating a struct instead of this pair might make things easier to read down the line

include/trossen_ai_robot_devices/trossen_ai_driver.hpp
```diff
@@ -0,0 +1,123 @@
...
+#include "trossen_sdk_utils/constants.hpp"
+
+
+namespace trossen_ai_robot_devices {
+    /**
+     * @brief Class representing a Trossen AI Robotic Arm
+     * This is used as a wrapper around the TrossenArmDriver from the libtrossen_arm library
+     * to provide a simplified interface for connecting, disconnecting, reading, writing,
+     * and staging the robotic arm.
+     * This was designed to be friendly for use in the LeRobot framework.
```
lukeschmitt-tr: You can probably get rid of this line

```suggestion
```

include/trossen_ai_robot_devices/trossen_ai_driver.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+        /**
+         * @brief Constructor for TrossenAIArm
+         * @param name Name of the robotic arm
+         * @param ip_address IP address of the robotic arm
+         * @param model Model of the robotic arm (e.g., "leader", "follower")
+         * Initializes the TrossenAIArm with the given parameters
+         * 
+         */
+        TrossenAIArm(const std::string& name, const std::string& ip_address, const std::string& model) : name_(name), ip_address_(ip_address), model_(model) {}
```
lukeschmitt-tr: Move the constructor definition into the source file (everything after `:`)

```suggestion
        TrossenAIArm(const std::string& name, const std::string& ip_address, const std::string& model);
```

include/trossen_ai_robot_devices/trossen_ai_driver.hpp
```diff
@@ -0,0 +1,123 @@
...
+        
+        /**
+         * @brief Move constructor and move assignment operator
+         * Defaulted to allow moving of TrossenAIArm instances
+         * This is safe and efficient as it transfers ownership of resources         
+         */
+        TrossenAIArm(TrossenAIArm&&) noexcept = default;
+        TrossenAIArm& operator=(TrossenAIArm&&) noexcept = default;
```
lukeschmitt-tr: Explain why

include/trossen_dataset/dataset.hpp
```diff
@@ -0,0 +1,444 @@
...
+        std::vector<float> std;
+        int count = 0;
+    };
+
+    /** @brief Data structure to hold the data for a single frame in an episode
+     * This structure holds the timestamp, observation state, action, and other relevant information for a single frame
+     */
+    struct FrameData
+    {   
+        /// @brief Timestamp in seconds
```
lukeschmitt-tr: Timestamp from when? Time from start? Unix epoch timestamp?

include/trossen_dataset/dataset.hpp
```diff
@@ -0,0 +1,444 @@
...
+                                  double fps = 30.0);
+
+        /**
+         * @brief Verify the dataset structure and metadata
+         * @return True if the dataset structure and metadata are valid, false otherwise
+         */
+        bool verify() const;
+
+        /** @brief Add a frame to the current episode
+         * @param frame FrameData object representing the frame to be added
```
lukeschmitt-tr: nit

```suggestion
        /** 
         * @brief Add a frame to the current episode
         * @param frame FrameData object representing the frame to be added
```

include/trossen_dataset/dataset.hpp
```diff
@@ -0,0 +1,444 @@
...
+         * @return Number of episodes as a size_t
+         */
+        size_t get_num_episodes() const
+        {
+            return episodes_buffer_.size();
+        }
+
+        /**
+         * @brief Get the image path for the dataset
+         * @return Image path as a string
```
lukeschmitt-tr: Absolute? Relative to the dataset root?

include/trossen_dataset/dataset.hpp
```diff
@@ -0,0 +1,444 @@
...
+
+        /**
+         * @brief Convert saved images to videos for each episode
+         * This function processes the images saved in the dataset and converts them into video files
+         */
+        void convert_to_videos() const;
+
+        /**
+         * @brief Get the number of existing episodes in the dataset
+         * @return Number of existing episodes as an int
```
lukeschmitt-tr: What is the difference between this and `get_num_episodes`?

include/trossen_sdk_utils/config_parser_utils.hpp
```diff
@@ -0,0 +1,122 @@
...
+ * @param json_path Path to the JSON configuration file
+ * @return Shared pointer to the loaded LeaderConfigBase object
+ * This function reads the JSON file, determines the type of leader robot configuration,
+ * and deserializes it into the appropriate derived class of LeaderConfigBase.
+ */
+inline std::shared_ptr<trossen_sdk_config::LeaderConfigBase> load_leader_config(const std::string& json_path) {
+
+    // Read the JSON file
+    std::ifstream file(json_path);
+    if (!file.is_open()) {
```
lukeschmitt-tr: Does this also check if the file exists?

include/trossen_sdk_utils/config_parser_utils.hpp
```diff
@@ -0,0 +1,122 @@
...
+        throw std::runtime_error("Failed to open file: " + json_path);
+    }
+
+    // Parse the JSON content
+    nlohmann::json j;
+    file >> j;
+
+    // Extract the type and create the appropriate config object
+    const std::string& type = j.at("type");
```
lukeschmitt-tr: What if type field is not specified?

include/trossen_sdk_utils/config_parser_utils.hpp
```diff
@@ -0,0 +1,122 @@
...
+    
+    // Determine the type of robot and create the appropriate instance
+    if (config.get_name() == "widowxai") {
+        auto leader_config = dynamic_cast<const trossen_sdk_config::WidowXRobotConfig&>(config);
+        return std::make_shared<trossen_ai_robot_devices::robot::TrossenAIWidowXRobot>(leader_config);
+    } else if (config.get_name() == "bimanual_widowxai") {
+        auto robot_config = dynamic_cast<const trossen_sdk_config::BimanualWidowXRobotConfig&>(config);
+        return std::make_shared<trossen_ai_robot_devices::robot::TrossenAIBimanualWidowXRobot>(robot_config);
+    } else {
+        throw std::runtime_error("Unknown robot type in configuration");
```
lukeschmitt-tr: print off the unknown name as done above

include/trossen_sdk_utils/config_parser_utils.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+    // Determine the type of leader and create the appropriate instance
+    if (config.get_name() == "widowxai_leader") {
+        auto leader_config = dynamic_cast<const trossen_sdk_config::WidowXLeaderConfig&>(config);
+        return std::make_shared<trossen_ai_robot_devices::teleoperator::TrossenAIWidowXLeader>(leader_config);
+    } else if (config.get_name() == "bimanual_widowxai_leader") {
+        auto leader_config = dynamic_cast<const trossen_sdk_config::BimanualWidowXLeaderConfig&>(config);
+        return std::make_shared<trossen_ai_robot_devices::teleoperator::TrossenAIBimanualWidowXLeader>(leader_config);
+    } else {
+        throw std::runtime_error("Unknown leader type in configuration");
```
lukeschmitt-tr: Print off unknown here too

include/trossen_sdk_utils/config_utils.hpp
```diff
@@ -0,0 +1,182 @@
...
+    std::string ip_address;
+    /// @brief Camera configurations for the robot
+    std::vector<CameraConfig> cameras;  // Camera configuration for the follower arm
+
+    /** @brief Get the name of the robot
+     * @return Name of the robot as a string
+     */
+    std::string get_name() const override {
+        return name;
+    }
```
lukeschmitt-tr: You can include this in the base class since you're implementing the same method in every child class

include/trossen_sdk_utils/config_parser_utils.hpp
```diff
@@ -0,0 +1,122 @@
...
+}
+
+/**
+ * @brief Load follower robot configuration from a JSON file
+ * @param json_path Path to the JSON configuration file
+ * @return Shared pointer to the loaded RobotConfigBase object
+ * This function reads the JSON file, determines the type of follower robot configuration,
+ * and deserializes it into the appropriate derived class of RobotConfigBase.
+ */
+inline std::shared_ptr<trossen_sdk_config::RobotConfigBase> load_follower_config(const std::string& json_path) {
```
lukeschmitt-tr: > follower

Inconsistent naming between "robot" and "follower"

include/trossen_sdk_utils/config_utils.hpp
```diff
@@ -0,0 +1,182 @@
...
+/**
+ * @brief Camera configuration structure
+ * This structure holds the configuration parameters for a camera including its name, serial number,
+ * frames per second, resolution, and whether to use depth.
+ */
+struct CameraConfig {
+    /// @brief Name of the camera
+    std::string name;
+    /// @brief Serial number of the camera
+    std::string serial;
```
lukeschmitt-tr: Thoughts on serial vs something like id or path?

Not all cameras will use serial numbers (OpenCV, etc.)

include/trossen_sdk_utils/config_utils.hpp
```diff
@@ -0,0 +1,182 @@
...
+ * This structure holds the configuration parameters for a Bimanual WidowX robot including its name,
+ * IP addresses for both arms, and camera configurations. It inherits from RobotConfigBase.
+ */
+struct WidowXRobotConfig : public RobotConfigBase {
+    /// @brief Name of the robot
+    std::string name;
+    /// @brief IP address of the robot
+    std::string ip_address;
+    /// @brief Camera configurations for the robot
+    std::vector<CameraConfig> cameras;  // Camera configuration for the follower arm
```
lukeschmitt-tr: Two comments on this property - remove the inline one

include/trossen_sdk_utils/config_utils.hpp
```diff
@@ -0,0 +1,182 @@
...
+
+
+/**
+ * @brief Configuration structure for WidowX robot
+ * This structure holds the configuration parameters for a WidowX robot including its name and IP address.
+ * It inherits from LeaderConfigBase.
+ */
+struct WidowXLeaderConfig : public LeaderConfigBase {
+    /// @brief Name of the leader robot
+    std::string name;
```
lukeschmitt-tr: Don't need name prop if the base class has it

include/trossen_sdk_utils/constants.hpp
```diff
@@ -0,0 +1,56 @@
...
+#define TROSSEN_SDK_UTILS_CONSTANTS_HPP 
+#include <filesystem>
+    
+namespace trossen_sdk {
+
+    /// @brief Default root path for dataset storage
+    const std::filesystem::path DEFAULT_ROOT_PATH = std::filesystem::path(std::getenv("HOME")) / ".cache" / "trossen_dataset_collection_sdk";
+
+    /// @brief Format string for leader robot configuration file path
+    const std::string LEADER_ROBOT_CONFIG_FORMAT = "../config/{}_leader.json";
```
lukeschmitt-tr: Are you assuming that the user is in the `build/scripts` directory here? Can we handle this better?

lukeschmitt-tr: Maybe do something like a configured file using cmake to get the location this is installed to?

include/trossen_sdk_utils/constants.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+    /// @brief Follower model identifier
+    const std::string FOLLOWER_MODEL = "follower";
+    
+    // Metadata related files
+
+    const std::string INFO_JSON = "info.json";
+    const std::string EPISODES_JSONL = "episodes.jsonl";
+    const std::string EPISODE_STATS_JSONL = "episode_stats.jsonl";
+    const std::string TASKS_JSONL = "tasks.jsonl";
```
lukeschmitt-tr: swap the order of these and other similar constants i.e. JSON_INFO or JSONL_TASKS for clarity

apache-arrow-apt-source-latest-noble.deb
lukeschmitt-tr, line 1: remove this file

include/trossen_sdk_utils/control_utils.hpp
```diff
@@ -0,0 +1,76 @@
...
+                    trossen_dataset::TrossenAIDataset& dataset,
+                    bool display_cameras,
+                    double fps);
+    /**
+     * @brief Control loop for teleoperated robot without dataset recording for teleoperation only
+     * @param robot Shared pointer to the robot object
+     * @param teleop_robot Shared pointer to the teleoperation robot object
+     * @param control_time Control loop time in seconds
+     * @param display_cameras Flag to indicate if camera images should be displayed
+     * @param fps Frames per second for video conversion
```
lukeschmitt-tr: > for video conversion

Doesn't seem to be the case for this method

src/trossen_ai_robot_devices/trossen_ai_cameras.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+    // Enable the device using the serial number
+    if (!serial_number_.empty()) {
+        cfg.enable_device(serial_number_);
+    }
+
+    // Enable the color stream as default
+    cfg.enable_stream(
+                RS2_STREAM_COLOR, capture_width_, capture_height_, RS2_FORMAT_RGB8, fps_
+            );
```
lukeschmitt-tr: style nit

```suggestion
    cfg.enable_stream(
        RS2_STREAM_COLOR, capture_width_, capture_height_, RS2_FORMAT_RGB8, fps_
```

src/trossen_ai_robot_devices/trossen_ai_cameras.cpp
```diff
@@ -0,0 +1,252 @@
...
+    
+    // Enable depth stream if use_depth_ is true
+    if (use_depth_) {
+        cfg.enable_stream(
+            RS2_STREAM_DEPTH, capture_width_, capture_height_, RS2_FORMAT_Z16, fps_
+        );
```
lukeschmitt-tr:
```suggestion
            RS2_STREAM_DEPTH, capture_width_, capture_height_, RS2_FORMAT_Z16, fps_);
```

include/trossen_ai_robot_devices/trossen_ai_cameras.hpp
```diff
@@ -0,0 +1,174 @@
...
+    cv::Mat depth_map;
+};
+
+
+/**
+ * @brief Class representing a Trossen AI Camera using Intel RealSense
+ * This class provides methods to connect to the camera, disconnect, and read frames
+ * It supports both color images and depth maps, and can be configured for resolution and frame rate
+ */
+class TrossenAICamera {
```
lukeschmitt-tr, line 49: This implementation assumes we'll only use realsense cameras - can we add an opencv option too? That way, we can abstract away the important interfaces and have two examples to build on in the future

src/trossen_ai_robot_devices/trossen_ai_cameras.cpp
```diff
@@ -0,0 +1,252 @@
...
+    try {
+        frames = camera_.wait_for_frames(3000); 
+    } catch (const rs2::error& e) {
+        spdlog::error("Failed to get frameset from camera: {}. Error: {}", name_, e.what());
+        return trossen_ai_robot_devices::ColorDepthData{};
+    }
+
+    // Extract color and depth frames
+    trossen_ai_robot_devices::ColorDepthData data;
+    if (frames && frames.size() > 0) {
```
lukeschmitt-tr: Can we do early return here since this block is already so indented?

```
if (fail) {
    log error
    return default
}

...normal logic...
```

src/trossen_ai_robot_devices/trossen_ai_cameras.cpp
```diff
@@ -0,0 +1,252 @@
...
+        // Get the timestamp for the frame
+        int64_t timestamp = color_and_depth.color_image.get_timestamp();
+        // Lock and set the result
+        {
+            std::lock_guard<std::mutex> lock(mtx);
+            if(use_depth_) {
+                result = trossen_ai_robot_devices::ImageData{name_, image, depth_map};
+            } else {
+                result = trossen_ai_robot_devices::ImageData{name_, image, cv::Mat()};
+            }
```
lukeschmitt-tr: depth_map is already a default cv::Mat we don't have to add this additional logic

src/trossen_ai_robot_devices/trossen_ai_cameras.cpp
```diff
@@ -0,0 +1,252 @@
...
+                std::filesystem::path parent_dir = file_path.parent_path();
+                if (!std::filesystem::exists(parent_dir)) {
+                    std::filesystem::create_directories(parent_dir);
+                }
+                if (!cv::imwrite(filename, bgr_image)) {
+                    spdlog::error("Failed to write color image to: {}", filename);
+                }
+            } catch (const cv::Exception& e) {
+                spdlog::error("Exception while writing color image to {}: {}", filename, e.what());
+            }
```
lukeschmitt-tr: I don't see a difference between this and the above block other than what's logged. Can the color/depth frame saving logic be combined?

src/trossen_ai_robot_devices/trossen_ai_cameras.cpp
```diff
@@ -0,0 +1,252 @@
...
+            } else {
+                depth_image = image;
+            }
+            try {
+                // Check if the parent directory exists, if not create it
+                std::filesystem::path file_path(filename);
+                std::filesystem::path parent_dir = file_path.parent_path();
+                if (!std::filesystem::exists(parent_dir)) {
+                    std::filesystem::create_directories(parent_dir);
+                }
```
lukeschmitt-tr: This directory creation feels wrong to put here - can this be moved to the dataset creation/init?

src/trossen_ai_robot_devices/trossen_ai_cameras.cpp
```diff
@@ -0,0 +1,252 @@
...
+                    spdlog::error("Failed to write color image to: {}", filename);
+                }
+            } catch (const cv::Exception& e) {
+                spdlog::error("Exception while writing color image to {}: {}", filename, e.what());
+            }
+        }
+    }
+}
+
+} // namespace trossen_data_collection_sdk
```
lukeschmitt-tr:
```suggestion
} // namespace trossen_ai_robot_devices
```

src/trossen_ai_robot_devices/trossen_ai_driver.cpp
```diff
@@ -0,0 +1,133 @@
...
+    if (!is_connected_) {
+        spdlog::error("Not connected to {}. Cannot write data.", name_);
+        return;
+    }
+    if (data_name == trossen_sdk::POSITION) {
+        // Check if any mode is not set to position
+        const auto& modes = driver_.get_modes();
+        if (!std::all_of(modes.begin(), modes.end(), [](const auto& mode){ return mode == trossen_arm::Mode::position; })) {
+            driver_.set_all_modes(trossen_arm::Mode::position);
+        }
```
lukeschmitt-tr: I think we should throw instead of setting - this may cause unexpected behavior for the user

src/trossen_ai_robot_devices/trossen_ai_driver.cpp
```diff
@@ -0,0 +1,133 @@
...
+        bool all_efforts = std::all_of(modes.begin(), modes.end(), [](const auto& mode){ return mode == trossen_arm::Mode::external_effort; });
+        if (!all_efforts) {
+            driver_.set_all_modes(trossen_arm::Mode::external_effort);
+        }
+        // Check if the data size is correct for external efforts
+        if (data.size() != driver_.get_num_joints()) {
+            spdlog::error("Invalid data size for external_efforts. Expected {} values.", driver_.get_num_joints());
+            return;
+        }
+        driver_.set_all_external_efforts(data, time_to_move_, false);
```
lukeschmitt-tr: Should we allow writing effort or velocity?

src/trossen_ai_robot_devices/trossen_ai_driver.cpp
```diff
@@ -0,0 +1,133 @@
...
+        return;
+    }
+    }
+    catch(const std::exception& e)
+    {
+        spdlog::error("Exception occurred: {}", e.what());
+    }
+    // Stage the arm
+    driver_.set_all_modes(trossen_arm::Mode::position);
+    driver_.set_all_positions({0.0, M_PI/3, M_PI/6, M_PI/5, 0.0, 0.0, 0.0}, 2.0, true);
```
lukeschmitt-tr: Looks like we can call stage_arm here

src/trossen_ai_robot_devices/trossen_ai_robot.cpp
```diff
@@ -0,0 +1,293 @@
...
+
+        TrossenAIWidowXRobot::TrossenAIWidowXRobot(const trossen_sdk_config::WidowXRobotConfig& config)
+            : name_(config.name), ip_address_(config.ip_address) {
+
+            robot_driver_ = std::make_unique<trossen_ai_robot_devices::TrossenAIArm>(config.name, config.ip_address, trossen_sdk::FOLLOWER_MODEL);
+            for (const auto& cam_config : config.cameras) {
+                cameras_.emplace_back(cam_config.name, cam_config.serial, cam_config.width, cam_config.height, cam_config.fps, cam_config.use_depth);
+            }
+            std::cout << "TrossenAIWidowXRobot initialized with name: " << name_
+                      << ", IP address: " << ip_address_ << std::endl;
```
lukeschmitt-tr: using cout instead of logging in much of this file

include/trossen_ai_robot_devices/trossen_ai_robot.hpp
```diff
@@ -0,0 +1,327 @@
...
+
+    class TrossenLeader{
+
+        public:
+            /** @brief Connect to the robot leader */
+            virtual void connect() = 0;
+            /** @brief Disconnect from the robot leader */
+            virtual void disconnect() = 0;
+            /** @brief Calibrate the robot leader */
+            virtual void calibrate() = 0;
```
lukeschmitt-tr: Do we need this?

include/trossen_ai_robot_devices/trossen_ai_robot.hpp
```diff
@@ -0,0 +1,327 @@
...
+             * @return Vector of joint feature names
+            */
+            virtual std::vector<std::string> get_joint_features() const = 0;
+
+            /** @brief Get the observation feature names of the robot 
+             * @return Vector of observation feature names
+            */
+            virtual std::vector<std::string> get_observation_features() const = 0;
+
+            /** @brief Get the camera names of the robot
```
lukeschmitt-tr: nit

```suggestion
            /**
             * @brief Get the camera names of the robot 
```

include/trossen_ai_robot_devices/trossen_ai_robot.hpp
```diff
@@ -0,0 +1,327 @@
...
+
+            /** @brief Get the observation feature names of the robot 
+             * @return Vector of observation feature names
+            */
+            virtual std::vector<std::string> get_observation_features() const = 0;
+
+            /** @brief Get the camera names of the robot 
+             * @return Vector of pairs containing camera name and type (e.g., "depth" or "color")
+            */
+            virtual std::vector<std::pair<std::string, std::string>> get_camera_names() const = 0;
```
lukeschmitt-tr: This can probably be defined in the base class as both overrides have the same implementation

lukeschmitt-tr: suggestion for function renaming

```suggestion
            virtual std::vector<std::pair<std::string, std::string>> get_camera_names_and_types() const = 0;
```

src/trossen_ai_robot_devices/trossen_ai_robot.cpp
```diff
@@ -0,0 +1,293 @@
...
+            state.observation_state = robot_driver_->read(trossen_sdk::POSITION);
+            // Read camera images 
+            for (auto& camera : cameras_) {
+                state.images.push_back(camera.async_read());
+            }
+            return state;
+        }
+
+        void TrossenAIWidowXRobot::send_action(const std::vector<double>& action) {
+            // Send joint position commands to the robot arm
```
lukeschmitt-tr: I don't think we need comments like this

include/trossen_ai_robot_devices/trossen_ai_robot.hpp
lukeschmitt-tr: <std::pair<std::string, std::string>

Can we make this a type/struct instead for usage clarity?

src/trossen_ai_robot_devices/trossen_ai_robot.cpp
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+        void TrossenAIBimanualWidowXRobot::configure() {
+            // Stage both arms to safe positions
+            right_robot_driver_->stage_arm();
+            left_robot_driver_->stage_arm();
+        }
+
+        trossen_ai_robot_devices::State TrossenAIBimanualWidowXRobot::get_observation() {
+            trossen_ai_robot_devices::State state;
```
lukeschmitt-tr: We create a lot of trossen_ai_robot_devices::State variables - both here and in the callers to this method. Can we change `get_observation` to accept a mutable reference and just populate it in this method?

src/trossen_ai_robot_devices/trossen_ai_robot.cpp
```diff
@@ -0,0 +1,293 @@
...
+            // Read camera images
+            for (auto& camera : cameras_) {
+                state.images.push_back(camera.async_read());
+            }
+            return state;
+        }
+
+        void TrossenAIBimanualWidowXRobot::send_action(const std::vector<double>& action) {
+            // Split the action vector and send commands to both robot arms
+            if (action.size() < right_robot_driver_->get_joint_names().size() + left_robot_driver_->get_joint_names().size()) {
```
lukeschmitt-tr: > get_joint_names().size()

We're creating a copy of the vector of joint names every time we call this method. Can we handle this count better? Maybe create a class property that keeps track of the number of action and observation features of the robot and just return that

src/trossen_ai_robot_devices/trossen_ai_robot.cpp
```diff
@@ -0,0 +1,293 @@
...
+            std::vector<std::string> right_features, left_features;
+            // Prefix joint names with "right_" and "left_" and suffix with ".pos"
+            // This is to allow compatibility with LeRobot replay and visualization tools
+            for (const auto& name : right_features_raw) {
+                right_features.push_back("right_" + name + ".pos");
+            }
+            for (const auto& name : left_features_raw) {
+                left_features.push_back("left_" + name + ".pos");
+            }
+            right_features.insert(right_features.end(), left_features.begin(), left_features.end());
```
lukeschmitt-tr: Why not just make `features` and push back the modified right_ left_ features?

src/trossen_ai_robot_devices/trossen_ai_robot.cpp
```diff
@@ -0,0 +1,293 @@
...
+            return camera_names;
+        }
+
+    }
+
+    namespace teleoperator {
+        
+        TrossenAIWidowXLeader::TrossenAIWidowXLeader(const trossen_sdk_config::WidowXLeaderConfig& config)
+            : name_(config.name), ip_address_(config.ip_address) {
+            robot_driver_ = std::make_unique<trossen_ai_robot_devices::TrossenAIArm>(config.name, config.ip_address, trossen_sdk::LEADER_MODEL);
```
lukeschmitt-tr: Lacking initialization logging

src/trossen_ai_robot_devices/trossen_ai_robot.cpp
```diff
@@ -0,0 +1,293 @@
...
+        void TrossenAIWidowXLeader::disconnect() {
+            robot_driver_->stage_arm(); // Stop the arm
+            robot_driver_->disconnect();
+            is_connected_ = false;
+            spdlog::info("Disconnected from leader: {}", name_);
+        }
+
+
+        TrossenAIBimanualWidowXLeader::TrossenAIBimanualWidowXLeader(const trossen_sdk_config::BimanualWidowXLeaderConfig& config)
+            : name_(config.name), left_ip_address_(config.left_ip_address), right_ip_address_(config.right_ip_address) {
```
lukeschmitt-tr: Initialization logging

src/trossen_ai_robot_devices/trossen_ai_robot.cpp
```diff
@@ -0,0 +1,293 @@
...
+            }
+            robot_driver_->connect();
+        }
+    
+
+        void TrossenAIWidowXLeader::calibrate() {
+        }
+
+        void TrossenAIWidowXLeader::configure() {
+            robot_driver_->stage_arm(); // Stage the arm to a safe position
```
lukeschmitt-tr: You can remove this comment

src/trossen_ai_robot_devices/trossen_ai_robot.cpp
```diff
@@ -0,0 +1,293 @@
...
+            std::vector<double> positions = robot_driver_->read(trossen_sdk::POSITION);
+            return positions;
+        }
+
+        void TrossenAIWidowXLeader::send_feedback() {
+
+        }
+
+        void TrossenAIWidowXLeader::disconnect() {
+            robot_driver_->stage_arm(); // Stop the arm
```
lukeschmitt-tr: Don't need this comment either

include/trossen_dataset/dataset.hpp
```diff
@@ -0,0 +1,444 @@
...
+         */
+        int get_existing_episodes() const;
```
lukeschmitt-tr: suggest get_num_existing_episodes or get_count_existing_episodes

src/trossen_dataset/dataset.cpp
```diff
@@ -0,0 +1,1137 @@
...
+
+        // Create episode folder name with zero-padded episode index
+        // TODO Use string formatting utility
+        std::string episode_folder_name = fmt::format("episode_{:06}", current_episode_->get_episode_idx());
+
+        // For each camera, create a folder for the episode if it doesn't exist
+        std::vector<std::pair<std::string, std::string>> camera_names = robot_->get_camera_names();
+        if (camera_names.empty()) {
+            spdlog::warn("No cameras found on the robot.");
+        }
```
lukeschmitt-tr: `camera_names` is not being used, rather we're using `image_data.camera_name`

src/trossen_dataset/dataset.cpp
```diff
@@ -0,0 +1,1137 @@
...
+
+        // Add the frame to the current episode
+        current_episode_->add_frame(frame);
+
+
+        // Create episode folder name with zero-padded episode index
+        // TODO Use string formatting utility
+        std::string episode_folder_name = fmt::format("episode_{:06}", current_episode_->get_episode_idx());
+
+        // For each camera, create a folder for the episode if it doesn't exist
```
lukeschmitt-tr: > create a folder for the episode

Is this comment accurate? I don't see us creating any directories

src/trossen_dataset/dataset.cpp
lukeschmitt-tr, line 1: Not really reviewing the loading/saving components of this just yet

src/trossen_sdk_utils/control_utils.cpp
```diff
@@ -0,0 +1,164 @@
...
+        }
+        
+        // Add frame to the dataset
+        dataset.add_frame(frame_data);
+
+        if (display_cameras) {
+            // Display images/videos using OpenCV
+            display_images(state.images);
+        }
+        // TODO: Use FPS to control the loop frequency
```
lukeschmitt-tr: Is this TODO done?

scripts/record.cpp
lukeschmitt-tr, line 1: not reviewing the scripts just yet

README.md
```diff
@@ -1 +1,257 @@
-# trossen_sdk
+# Trossen Robotics Data Collection SDK
+
+This software was tested on Ubuntu 24.04 and the installation instructions were teseted on the TOTL PC fresh ISO image.
```
lukeschmitt-tr: This line should not be first - the description of this software should

Also, assume that customers do not have the TOTL PC - this will be available to everyone

README.md
```diff
@@ -1 +1,257 @@
...
+sudo apt install -y -V libarrow-dataset-glib-dev # For Apache Arrow Dataset GLib (C)
+sudo apt install -y -V libarrow-acero-dev # For Apache Arrow Acero
+sudo apt install -y -V libarrow-flight-dev # For Apache Arrow Flight C++
+sudo apt install -y -V libarrow-flight-glib-dev # For Apache Arrow Flight GLib (C)
+sudo apt install -y -V libarrow-flight-sql-dev # For Apache Arrow Flight SQL C++
+sudo apt install -y -V libarrow-flight-sql-glib-dev # For Apache Arrow Flight SQL GLib (C)
+sudo apt install -y -V libgandiva-dev # For Gandiva C++
+sudo apt install -y -V libgandiva-glib-dev # For Gandiva GLib (C)
+sudo apt install -y -V libparquet-dev # For Apache Parquet C++
+sudo apt install -y -V libparquet-glib-dev # For Apache Parquet GLib (C)
```
lukeschmitt-tr: Install everything in one call

```suggestion
sudo apt install -y -V \
  libarrow-dev \
  libarrow-glib-dev \
  libarrow-dataset-dev \
  libarrow-dataset-glib-dev \
  libarrow-acero-dev \
  libarrow-flight-dev \
  libarrow-flight-glib-dev \
  libarrow-flight-sql-dev \
  libarrow-flight-sql-glib-dev \
  libgandiva-dev \
  libgandiva-glib-dev \
  libparquet-dev \
  libparquet-glib-dev
```

README.md
```diff
@@ -1 +1,257 @@
...
+--overwrite false \
+--dataset test_dataset_00 \
+--root ~/.cache/huggingface/lerobot/ \
+--repo_id trossen-ai/trossen-widowx \
+--single_task pick_place \
+--num_image_writer_threads_per_camera 4 \
+--num_image_writer_processes 1 \
+--video true \
+--run_compute_stats true
+```
```
lukeschmitt-tr:
```suggestion
```bash
./record \
  --robot_name trossen_ai_stationary \
  --recording_time 10 \
  --num_episodes 1 \
  --fps 30 \
  --display_cameras true \
  --tags test \
  --overwrite false \
  --dataset test_dataset_00 \
  --root ~/.cache/huggingface/lerobot/ \
  --repo_id trossen-ai/trossen-widowx \
  --single_task pick_place \
  --num_image_writer_threads_per_camera 4 \
  --num_image_writer_processes 1 \
  --video true \
  --run_compute_stats true
```
```

File: README.md [hunk 54]
```diff
@@ -1 +1,257 @@
...
+|    0   N/A  N/A            2620      G   /usr/bin/gnome-shell                     56MiB |
+|    0   N/A  N/A            3568      G   ...exec/xdg-desktop-portal-gnome          9MiB |
+|    0   N/A  N/A            5273      G   ...slack/212/usr/lib/slack/slack         51MiB |
+|    0   N/A  N/A            5798      G   ...85b68b8157e8c917bdaf1308fa936        111MiB |
+|    0   N/A  N/A           38204      G   /proc/self/exe                           93MiB |
+|    0   N/A  N/A           48960      G   /usr/bin/gnome-control-center            21MiB |
+|    0   N/A  N/A           79911      G   /usr/bin/nautilus                        20MiB |
++-----------------------------------------------------------------------------------------+
+```
+
```
lukeschmitt-tr: Missing docs on building the sdk itself

README.md
```diff
@@ -1 +1,257 @@
...
+To view the dataset online in the HuggingFace dataset viewer, you can use the following link:
+
+[LeRobot Dataset Viewer](https://huggingface.co/spaces/lerobot/visualize_dataset)
+
+Just dataset name and repo id in the input box and click on "Go".
+
+
+## Put Arms to Sleep
+
+In any case where the system crashes or you need to quickly disable the motors, you can use the `sleep` script to safely disconnect from the robot.
```
lukeschmitt-tr: Description seems more frantic than needed

README.md
```diff
@@ -1 +1,257 @@
...
+./sleep --robot trossen_ai_stationary
+```
+
+## Teleoperation
+
+If you want to do a dry run of your experiment without recording, you can use the teleop script to control the robot.
+
+```bash
+./teleop --robot trossen_ai_stationary --fps 30
+```
```
lukeschmitt-tr: Put our scripts first - add a LeRobot compatibility section that contains the viz, replay, and upload sections

lukeschmitt-tr (changes requested):

README.md
shantanuparab-tr: Ohh, I was thinking this was for internal use

include/trossen_ai_robot_devices/trossen_ai_robot.hpp
shantanuparab-tr: For later "SO-101"

src/trossen_ai_robot_devices/trossen_ai_driver.cpp
shantanuparab-tr: Yes

src/trossen_ai_robot_devices/trossen_ai_driver.cpp
shantanuparab-tr: This was done to avoid explicitly setting the modes before calling read and write methods.

include/trossen_sdk_utils/config_utils.hpp
shantanuparab-tr: I am using unique_identifier and then parse it appropriately in respective classes

include/trossen_sdk_utils/config_parser_utils.hpp
shantanuparab-tr: No, added a check

src/trossen_ai_robot_devices/trossen_ai_cameras.cpp
shantanuparab-tr: If I initialize this in dataset creation, I have to specify how many episodes I need ahead of time. Cause images are stored in their respective episode folders. Even if I do that, there will be unused folders in case the controller terminates early.

shantanuparab-tr: Let me think where this will make more sense, maybe episode initialization.

include/trossen_dataset/dataset.hpp
shantanuparab-tr: Get existing episodes is used for resuming, it will check the directory and fill up the episode buffer. get num episodes will give the size of the episode buffer and is used to keep track of episode in the current recording session.

include/trossen_ai_robot_devices/trossen_ai_robot.hpp
shantanuparab-tr: Moving this fucntion to the base will require me to move the cameras_ member variable to base as well.

I am removing it from the base class and the user can implement the method if they need. As robot's can exist without any cameras as well. I'll add a check for cameras before invoking any camera based code.

include/trossen_ai_robot_devices/trossen_ai_cameras.hpp
```diff
@@ -0,0 +1,220 @@
...
+   * Logs that the camera is being disconnected
+   */
+  virtual void disconnect() = 0;
+
+  /**
+   * @brief Read a frame from the camera
+   * Waits for a new frameset and extracts color and depth frames
+   * @return ImageData structure containing color image and depth map frames
+   */
+  virtual trossen_ai_robot_devices::ImageData read() = 0;
```
lukeschmitt-tr: This should be an out-param to avoid copies

lukeschmitt-tr: we're already in this namespace

```suggestion
  virtual ImageData read() = 0;
```

include/trossen_ai_robot_devices/trossen_ai_cameras.hpp
```diff
@@ -0,0 +1,220 @@
...
+
+  /**
+   * @brief Asynchronously read a frame from the camera
+   * @return ImageData structure containing color image and depth map
+   */
+  virtual ImageData async_read() = 0;
```
lukeschmitt-tr: out-param

include/trossen_ai_robot_devices/trossen_ai_driver.hpp
```diff
@@ -0,0 +1,138 @@
...
+ */
+class TrossenAIArm {
+ public:
+  /**
+   * @brief Constructor for TrossenAIArm
+   * @param name Name of the robotic arm
+   * @param ip_address IP address of the robotic arm
+   * @param model Model of the robotic arm (e.g., "leader", "follower")
+   * Initializes the TrossenAIArm with the given parameters
+   *
```
lukeschmitt-tr: nit

```suggestion
```

include/trossen_ai_robot_devices/trossen_ai_robot.hpp
```diff
@@ -0,0 +1,330 @@
...
+struct State {
+  /// @brief Observation state of the robot (e.g., joint positions)
+  std::vector<double> observation_state;
+  /// @brief Action to be taken by the robot
+  std::vector<double> action;
+  /// @brief Images captured for the frame
+  std::vector<ImageData> images;
+};
+
+struct CameraType {
```
lukeschmitt-tr: needs struct-level docstring

src/trossen_ai_robot_devices/trossen_ai_cameras.cpp
```diff
@@ -0,0 +1,440 @@
...
+}
+
+void RealsenseCamera::connect() {
+  // Create a realsense config
+  rs2::config cfg;
+
+  // Enable the device using the unique ID
+  if (!unique_id_.empty()) {
+    cfg.enable_device(unique_id_);
+  }
```
lukeschmitt-tr, line 59: What happens if we don't set the unique_id_? What happens if we fail to enable the device?

src/trossen_ai_robot_devices/trossen_ai_cameras.cpp
```diff
@@ -0,0 +1,440 @@
...
+    frames = camera_.wait_for_frames(3000);
+  } catch (const rs2::error& e) {
+    spdlog::error("Failed to get frameset from camera: {}. Error: {}", name_,
+                  e.what());
+    return trossen_ai_robot_devices::ImageData{};
+  }
+
+  // Extract color and depth frames
+  trossen_ai_robot_devices::ImageData data;
+  data.camera_name = name_;
```
lukeschmitt-tr: We can initialize the ImageData struct before the try/catch and still return the empty struct.

src/trossen_ai_robot_devices/trossen_ai_cameras.cpp
```diff
@@ -0,0 +1,440 @@
...
+  // Try to get color frame
+  try {
+    rs2::frame color_frame = frames.get_color_frame();
+
+    if (color_frame) {
+      // Convert rs2::frame to cv::Mat
+      const void* data_ptr = static_cast<const void*>(color_frame.get_data());
+      cv::Mat image(cv::Size(capture_width_, capture_height_), CV_8UC3,
+                    const_cast<void*>(data_ptr), cv::Mat::AUTO_STEP);
+      data.image = image;
```
lukeschmitt-tr: `data.image = image` is an assignment operation, not a copy operation, right? We're just assigning header metadata and passing the data pointer?

include/trossen_ai_robot_devices/trossen_ai_cameras.hpp
lukeschmitt-tr: Is this doing a copy operation?

src/trossen_ai_robot_devices/trossen_ai_cameras.cpp
```diff
@@ -0,0 +1,440 @@
...
+  // Note: OpenCV does not natively support depth maps, so this is left empty
+  if (use_depth_) {
+    spdlog::warn("Depth map requested but not supported in OpenCVCamera: {}",
+                 name_);
+  }
+
+  return data;
+}
+
+trossen_ai_robot_devices::ImageData OpenCVCamera::async_read() {
```
lukeschmitt-tr: This looks like it can probably be moved to the base class if this is doing the same thing for both supported camera types?

src/trossen_ai_robot_devices/trossen_ai_robot.cpp
```diff
@@ -0,0 +1,372 @@
...
+  left_robot_driver_->connect();
+  for (auto& camera : cameras_) {
+    camera->connect();
+  }
+  is_connected_ = true;
+}
+
+void TrossenAIBimanualWidowXRobot::calibrate() {
+  // TODO(shantanuparab-tr) Implement calibration logic if needed
+  // Compliant Gripper Calibration
```
lukeschmitt-tr: >   // Compliant Gripper Calibration

Does this need to be here?

.gitignore
```diff
@@ -0,0 +1,4 @@
+/build
+/.vscode
+.clang-format
```
lukeschmitt-tr: You can commit the `.clang-format` file - we want to share linter configurations

lukeschmitt-tr (changes requested):

src/trossen_ai_robot_devices/trossen_ai_cameras.cpp
shantanuparab-tr: throwing an error here

src/trossen_ai_robot_devices/trossen_ai_cameras.cpp
shantanuparab-tr: I am making this into a move operation to transfer ownership of image, it was a copying the image initially.

docs/readme/configuration.md
```diff
@@ -0,0 +1,110 @@
...
+For Realsense Cameras:
+- Use the realsense-viewer application to find the serial number of your camera. More details [here](https://docs.trossenrobotics.com/trossen_arm/main/tutorials/lerobot/configuration.html#camera-unique_identifier-number).
+
+For OpenCV Cameras:
+- On first run, the program will scan through all available cameras and store images from the cameras in the output folder along with their indices. You can use this index to configure the camera in the configuration file.
+- You can check the output folder to match camera views with their corresponding index and update the configuration file accordingly.
+
+## Leader Robot Configuration
+
+Below is an example configuration for a `bimanual WidowXAI Leader` robot using  `RealSense` cameras:
```
lukeschmitt-tr: no cameras on leaders

docs/readme/installation.md
```diff
@@ -0,0 +1,179 @@
...
+
+```bash
+sudo apt update
+sudo apt install libopencv-dev
+```
+
+## Nvidia Driver
+
+If you are using a system with an Nvidia GPU, you may need to install the appropriate Nvidia driver for your GPU model.
+This does not play any role in the SDK functionality but is required for the realsense-viewer to work properly. This drivers will be used to display the camera streams using the realsense-viewer.
```
lukeschmitt-tr, line 105: Is this true?

docs/readme/lerobot.md
```diff
@@ -0,0 +1,56 @@
...
+```bash
+ huggingface-cli upload TrossenRoboticsCommunity/record-test ~/.cache/huggingface/lerobot/TrossenRoboticsCommunity/test_dataset_03 --repo-type dataset --revision v2.1
+```
+
+The above command uploads all files including images.
+
+
+## Replaying using LeRobot
+
+If you have `LeRobot` frame work installed, you can also replay using the replay scripts.
```
lukeschmitt-tr:
```suggestion
If you have the `LeRobot` framework installed, you can also replay using the replay scripts.
```

docs/readme/lerobot.md
```diff
@@ -0,0 +1,56 @@
...
+Note: This was tested using the new api integration for Trossen Arms with LeRobot.
```
lukeschmitt-tr, line 32: > new api integration

elaborate - maybe link to our docs

docs/readme/usage.md
```diff
@@ -0,0 +1,102 @@
...
+```bash
+make run-record ARGS="\
+  --robot trossen_ai_stationary \
+  --recording_time 10 \
+  --num_episodes 1 --fps 30 \
+  --display_cameras true \
+  --tags test \
+  --overwrite true \
+  --dataset test_dataset_00 \
+  --root ~/.cache/huggingface/lerobot/ \
```
lukeschmitt-tr: Do we usually save to this location, or is this just for lerobot compatibility and for alternate usage?

docs/readme/usage.md
```diff
@@ -0,0 +1,102 @@
+# Usage
+
+
+You can run the following commands from the root of the repository after building the project.
+
+## Recording a dataset
+
+```bash
+make run-record ARGS="\
```
lukeschmitt-tr: Let's just use the program binary directly instead of using the makefile

src/trossen_ai_robot_devices/trossen_ai_robot.cpp
```diff
@@ -88,7 +88,9 @@ void TrossenAIWidowXRobot::get_observation(
   state->observation_state = robot_driver_->read(trossen_sdk::POSITION);
   // Read camera images
   for (auto& camera : cameras_) {
-    state->images.push_back(camera->async_read());
+    trossen_ai_robot_devices::ImageData img_data;
```
lukeschmitt-tr: This constructor can be moved outside of the loop to conserve resources.

1. We build an empty imagedata
2. We fill an imagedata with data - this allocates memory to store the data
3. We can then reuse the pre-allocated memory to store new data

src/trossen_ai_robot_devices/trossen_ai_cameras.cpp
```diff
@@ -183,24 +199,28 @@ OpenCVCamera::OpenCVCamera(const std::string& name,
...
-    return;
+    find_cameras();
+    spdlog::info(
+      "Use the index from the available cameras listed above as the unique_id."
+      "Please check outputs folder to get images associated with each camera.");
+    throw std::runtime_error("Invalid unique_id for OpenCVCamera: " + unique_id_);
   }
 
-  cv::VideoCapture cap(device_index);
+  cv::VideoCapture cap(device_index, cv::CAP_V4L2);  // Using V4L2 backend on Linux
```
lukeschmitt-tr, line 223: What's the reasoning for this addition?

lukeschmitt-tr (changes requested):

docs/readme/installation.md
shantanuparab-tr: Yes, if the nvidia drivers are not setup correctly Realsense viewer fails to open its UI. And the user won't be able to find the serial numbers for the cameras.

lukeschmitt-tr (approved):

PR #197, opened 2026-03-18
Add depth camera awareness to sanity check

shantanuparab-tr:
SanityCheckConfig now includes depth_camera_producers to track cameras with
depth enabled. The sanity check reports a breakdown of color-only vs
color+depth records and expected MCAP backend writes (depth cameras produce
2 messages per record: 1 color + 1 depth).

The record count validation is unchanged since the sink counts each
ImageRecord as 1 regardless of embedded depth. The split exists for
diagnostic output to help debug depth frame loss.

Updated all example call sites (solo, stationary, mobile, so101) to compute
and pass the depth camera count. Also includes zed_camera in the camera FPS
detection loop.

PR #196, opened 2026-03-18
Add ZED camera hardware component and push producer

shantanuparab-tr:
Adds ZED SDK 5.x camera support as an opt-in build feature (TROSSEN_ENABLE_ZED):

- ZedCameraComponent: hardware component wrapping sl::Camera with full SDK 5.x
  InitParameters and RuntimeParameters support. Includes resolution/depth mode
  parsing with deprecation warnings for legacy CV-based modes (PERFORMANCE,
  QUALITY, ULTRA). Warmup loop settles auto-exposure before recording.

- ZedPushProducer: push-style producer using VIEW::LEFT_BGR for zero-conversion
  BGR retrieval (SDK 5.x). Depth maps are sanitized (NaN/Inf to 0) and converted
  from float32 meters to uint16 millimeters matching the RealSense Z16 convention.

- CMake integration: TROSSEN_ENABLE_ZED option (OFF by default), conditional
  compilation, find_package(ZED) + find_package(CUDA).

PR #195, opened 2026-03-18
Add extra field passthrough and numeric serial support to CameraConfig

shantanuparab-tr:
The CameraConfig struct now captures unknown JSON keys into an `extra` field
and merges them back in to_json(), enabling transparent passthrough of
hardware-specific parameters (e.g. ZED resolution, depth_mode, warmup_seconds)
without bloating the shared config struct.

Also handles numeric serial numbers (ZED uses integers) alongside the existing
string format (RealSense), fixing a json type_error.302 crash.

PR #198, opened 2026-03-23
Add lifecycle callbacks to SessionManager with deadlock fix

shantanuparab-tr:
Introduce four observer-pattern lifecycle hooks (on_pre_episode,
on_episode_started, on_episode_ended, on_pre_shutdown) and fix the
episode_mutex_ deadlock by splitting stats() into a public locking
wrapper and private stats_unlocked() for internal callers.

- Pre-episode callbacks can abort episode start (return false or throw)
- Shutdown callbacks fire exactly once (idempotent guard)
- All callbacks are exception-isolated
- Legacy set_episode_complete_callback() deprecated, delegates to on_episode_ended()
- 14 new unit tests (SM-10 re-enabled, SM-12 through SM-17)
- Two new examples: callback_demo and mock_producer_demo (no hardware required)


shantanuparab-tr:
* **#200** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/200?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#199** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/199?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#198** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/198?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/198?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#197** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/197?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#196** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/196?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#195** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/195?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

examples/mock_producer_demo/mock_producer_demo.cpp
lukeschmitt-tr, line 1: do we need this?

include/trossen_sdk/runtime/session_manager.hpp
```diff
@@ -204,11 +204,76 @@ class SessionManager {
...
+  using PreShutdownCallback = std::function<void()>;
+
+  /**
+   * @brief Register a callback invoked before each episode starts
+   *
+   * Called in start_episode() after backend/sink are ready but before the
+   * scheduler begins polling. If the callback returns false or throws, the
+   * episode is aborted and start_episode() returns false.
+   *
+   * Multiple callbacks are invoked in registration order.
```
lukeschmitt-tr, line 223: Is this true? Is there a demo for this?

include/trossen_sdk/runtime/session_manager.hpp
```diff
@@ -204,11 +204,76 @@ class SessionManager {
...
+  void on_pre_episode(PreEpisodeCallback cb);
+
+  /**
+   * @brief Register a callback invoked after each episode has started
+   *
+   * Called at the end of start_episode(), after the scheduler is running
+   * and episode_active_ is true.
+   *
+   * @param cb Callback to register
+   * @throws std::runtime_error if called while an episode is active
```
lukeschmitt-tr, line 239: called after episode_active_ is true but throws if episode started?

include/trossen_sdk/runtime/session_manager.hpp
```diff
@@ -377,6 +448,11 @@ class SessionManager {
   std::shared_ptr<io::Backend> create_backend(
     const ProducerMetadataList& producer_metadatas);
 
+  /**
+   * @brief Compute stats without acquiring episode_mutex_ (caller must hold it)
+   */
+  Stats stats_unlocked() const;
+
```
lukeschmitt-tr, line 443: can we just fix the current method instead of making a new one?

include/trossen_sdk/runtime/session_manager.hpp
```diff
@@ -204,11 +204,76 @@ class SessionManager {
...
+   * @param cb Callback to register
+   * @throws std::runtime_error if called while an episode is active
+   */
+  void on_episode_ended(EpisodeEndedCallback cb);
+
+  /**
+   * @brief Register a callback invoked at the top of shutdown()
+   *
+   * Called before stop_episode() during shutdown. Useful for returning
+   * hardware to safe positions.
```
lukeschmitt-tr: Does it make sense to have this be done before the episode stops? Are we still recording when this fires?

include/trossen_sdk/runtime/session_manager.hpp
```diff
@@ -204,11 +204,76 @@ class SessionManager {
...
+  using PreEpisodeCallback = std::function<bool()>;
+
+  /// @brief Callback invoked after an episode has fully started.
+  using EpisodeStartedCallback = std::function<void()>;
+
+  /// @brief Callback invoked after an episode has fully stopped.
+  using EpisodeEndedCallback = std::function<void(const Stats&)>;
+
+  /// @brief Callback invoked at the top of shutdown(), before stop_episode().
+  using PreShutdownCallback = std::function<void()>;
```
lukeschmitt-tr, line 214: Why do we want to differentiate between bool and void here?

lukeschmitt-tr: I guess with a pre-episode failure, you can handle it, but during or after, you really can't?

src/runtime/session_manager.cpp
```diff
@@ -195,6 +206,33 @@ bool SessionManager::start_episode() {
     }
   }
 
+  // Invoke pre-episode callbacks (after sink ready, before scheduler)
+  for (const auto& cb : pre_episode_cbs_) {
```
lukeschmitt-tr, line 216: Lots of duplication with the sink and backend memory management both below and above. Can we simplify this at all?

src/runtime/session_manager.cpp
```diff
@@ -383,10 +431,9 @@ SessionManager::Stats SessionManager::stats() const {
         s.remaining = std::chrono::duration<double>(0);
       }
     } else {
-      s.remaining = std::nullopt;  // unlimited
+      s.remaining = std::nullopt;
     }
 
-    // Compute records written to current episode as delta from start
     if (current_sink_) {
```
lukeschmitt-tr: Why remove these comments?

include/trossen_sdk/runtime/session_manager.hpp
```diff
@@ -204,11 +204,76 @@ class SessionManager {
...
+  // =========================================================================
+
+  /// @brief Callback invoked before an episode starts. Return false to abort.
+  using PreEpisodeCallback = std::function<bool()>;
+
+  /// @brief Callback invoked after an episode has fully started.
+  using EpisodeStartedCallback = std::function<void()>;
+
+  /// @brief Callback invoked after an episode has fully stopped.
+  using EpisodeEndedCallback = std::function<void(const Stats&)>;
```
lukeschmitt-tr, line 211: What's the difference between episode ended and the existing episode complete?

tests/test_session_manager.cpp
```diff
@@ -205,26 +205,15 @@ TEST_F(SessionManagerTest, PolledProducer_RecordsReachBackend) {
...
-// same thread. This affects ANY use of set_episode_complete_callback().
-//
-// Recommended fix: either (a) release episode_mutex_ before calling the callback,
-// or (b) use a separate internal stats computation that does not acquire the lock,
-// or (c) switch episode_mutex_ to std::recursive_mutex.
-//
-// This test is DISABLED until the deadlock is fixed.
-TEST_F(SessionManagerTest, DISABLED_EpisodeCompleteCallback_Invoked) {
+// SM-10: Episode complete callback (deadlock fixed via stats_unlocked)
+// ============================================================================
```
lukeschmitt-tr: Do we need these banner headers here? Can we just have comments describing the test case?

lukeschmitt-tr (changes requested):

examples/mock_producer_demo/mock_producer_demo.cpp
shantanuparab-tr: Removed. It was useful for the demo however, I'll just add updated SessionManager to the actual scripts and cleanup the test residues in the next PR.

include/trossen_sdk/runtime/session_manager.hpp
shantanuparab-tr: Yes — set_episode_complete_callback delegates to on_episode_ended() (session_manager.cpp:469-471). The callback_demo example demonstrates the new lifecycle API. I'll create a demo that shows the deprecation warning for internal use.

include/trossen_sdk/runtime/session_manager.hpp
shantanuparab-tr: Fixed the confusing wording — clarified that registration must happen before an episode starts, and removed the internal episode_active_ reference.

include/trossen_sdk/runtime/session_manager.hpp
shantanuparab-tr: stats_unlocked() exists to avoid a deadlock: stop_episode() already holds episode_mutex_ when it needs to compute final stats for the callback. Calling the locking stats() from there would deadlock. This is the standard lock-splitting pattern — cleaner than switching to a recursive mutex.

include/trossen_sdk/runtime/session_manager.hpp
shantanuparab-tr: Good point — changed the order so pre-shutdown callbacks fire after stop_episode(). Arms can go to sleep while post-processing/encoding runs.

include/trossen_sdk/runtime/session_manager.hpp
shantanuparab-tr: PreEpisodeCallback returns bool so it can abort the episode start (return false = don't start). The others are notification-only — nothing to abort after an episode has already started or ended.

shantanuparab-tr: Exactly right — pre-episode is the only point where aborting makes sense.

src/runtime/session_manager.cpp
shantanuparab-tr: Fixed — extracted a cleanup_and_abort lambda to deduplicate the three identical teardown blocks.

src/runtime/session_manager.cpp
shantanuparab-tr: Restored — the comments are back.

include/trossen_sdk/runtime/session_manager.hpp
shantanuparab-tr: They're the same — set_episode_complete_callback is the legacy API, now deprecated and delegating to on_episode_ended(). Kept for backward compatibility.

tests/test_session_manager.cpp
shantanuparab-tr: I'll remove the banners.

lukeschmitt-tr (approved):

shantanuparab-tr:
### Merge activity

* **Mar 27, 11:52 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/198).
* **Mar 27, 11:58 PM UTC**: [Graphite](https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/198) rebased this pull request as part of a merge.
* **Mar 27, 11:59 PM UTC**: @shantanuparab-tr merged this pull request with [Graphite](https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/198).

PR #204, opened 2026-03-31
Add discard_episode() to backend interface for re-record support

shantanuparab-tr:
When an operator messes up a demonstration they need to be able to throw it away and re-record at the same episode index. The backends currently only know how to finalize an episode (flush, encode videos, write stats). They have no way to delete one.

This adds a pure virtual discard_episode() to the Backend base class. It's pure virtual rather than a default no-op because a backend that silently does nothing on discard would leave stale files on disk and corrupt the dataset. A compile error for a missing override is safer.

Both backends extract their shared teardown logic into a private close_resources() method that close() and discard_episode() both call, avoiding the duplication that would otherwise exist between the two paths.

LeRobotV2Backend cleans up everything the episode could have written: the parquet file, image directories, encoded video files, and the metadata entries in episodes.jsonl, episodes_stats.jsonl, and info.json. This matters because discard can be called during the reset phase after close() has already run, meaning videos have been encoded and metadata has been appended. The frame count is read from the episodes.jsonl entry before removing it so info.json counters can be decremented accurately.

TrossenMCAPBackend closes the writer and deletes the single .mcap file. Path is reconstructed from config so it works even if the backend was already closed by the sink.

NullBackend is a no-op since there are no files to delete.

All implementations are safe to call whether the backend is still open or was already closed.

### DIFF
--- a/include/trossen_sdk/io/backends/lerobot_v2/lerobot_v2_backend.hpp
+++ b/include/trossen_sdk/io/backends/lerobot_v2/lerobot_v2_backend.hpp
@@ -418,6 +418,92 @@ inline bool write_task_entry(
 }
 
 /**
+ * @brief Remove the last non-empty line from a JSONL file
+ *
+ * Used during episode discard to undo appended entries.
+ *
+ * @param file_path Path to the JSONL file
+ * @return true on success, false on failure
+ */
+inline bool remove_last_jsonl_line(const std::filesystem::path& file_path) {
+  namespace fs = std::filesystem;
+  if (!fs::exists(file_path)) return true;
+
+  std::ifstream in(file_path);
+  if (!in.is_open()) return false;
+
+  std::vector<std::string> lines;
+  std::string line;
+  while (std::getline(in, line)) {
+    if (!line.empty()) lines.push_back(line);
+  }
+  in.close();
+
+  if (lines.empty()) return true;
+  lines.pop_back();
+
+  std::ofstream out(file_path, std::ios::trunc);
+  if (!out.is_open()) return false;
+  for (const auto& l : lines) {
+    out << l << "\n";
+  }
+  out.close();
+  return true;
+}
+
+/**
+ * @brief Revert info.json counters after discarding an episode
+ *
+ * Decrements total_episodes, total_frames, total_videos, and adjusts
+ * the train split range. Requires the frame count that was recorded for
+ * the discarded episode.
+ *
+ * @param meta_dir Path to the meta directory
+ * @param episode_frame_count Number of frames in the discarded episode
+ * @param num_videos Number of video streams in the discarded episode
+ * @return true on success, false on failure
+ */
+inline bool revert_info_json(
+    const std::filesystem::path& meta_dir,
+    int episode_frame_count,
+    int num_videos) {
+  namespace fs = std::filesystem;
+
+  fs::path info_path = meta_dir / JSON_INFO;
+  if (!fs::exists(info_path)) return true;
+
+  nlohmann::ordered_json info_json;
+  {
+    std::ifstream info_file(info_path);
+    if (!info_file.is_open()) return false;
+    info_file >> info_json;
+  }
+
+  int total_episodes = std::max(0, info_json.value("total_episodes", 0) - 1);
+  info_json["total_episodes"] = total_episodes;
+  info_json["total_videos"] = std::max(0, info_json.value("total_videos", 0) - num_videos);
+  info_json["total_frames"] = std::max(0, info_json.value("total_frames", 0) - episode_frame_count);
+
+  int chunks_size = info_json.value("chunks_size", 1000);
+  info_json["total_chunks"] = std::max(1, (total_episodes + chunks_size - 1) / chunks_size);
+
+  // Adjust train split end
+  std::string train_split = info_json["splits"].value("train", "0:0");
+  size_t colon_pos = train_split.find(':');
+  if (colon_pos != std::string::npos) {
+    int train_start = std::stoi(train_split.substr(0, colon_pos));
+    int train_end = std::max(train_start, std::stoi(train_split.substr(colon_pos + 1)) - 1);
+    info_json["splits"]["train"] = std::to_string(train_start) + ":" + std::to_string(train_end);
+  }
+
+  std::ofstream out(info_path);
+  if (!out.is_open()) return false;
+  out << info_json.dump(4);
+  out.close();
+  return true;
+}
+
+/**
  * @brief Append episode statistics to episodes_stats.jsonl
  *
  * @param meta_dir Path to the meta directory
@@ -625,6 +711,11 @@ public:
   void close() override;
 
   /**
+   * @brief Discard episode data and delete parquet file + image directories
+   */
+  void discard_episode() override;
+
+  /**
    * @brief Add metadata to be written to info.json
    *
    * @param md Metadata to add
@@ -761,6 +852,14 @@ public:
   }
 
 private:
+  /**
+   * @brief Close writer, output stream, and image workers without finalization.
+   *
+   * Shared teardown used by both close() and discard_episode(). Does NOT call
+   * convert_to_videos() or compute_statistics(). Caller must hold open_mutex_.
+   */
+  void close_resources();
+
   /**
    * @brief Write a joint state record to disk
    *
--- a/include/trossen_sdk/io/backends/lerobot_v2/lerobot_v2_backend.hpp
+++ b/include/trossen_sdk/io/backends/lerobot_v2/lerobot_v2_backend.hpp
@@ -418,14 +418,19 @@ inline bool write_task_entry(
 }
 
 /**
- * @brief Remove the last non-empty line from a JSONL file
+ * @brief Remove the last JSONL entry matching a given episode_index
  *
- * Used during episode discard to undo appended entries.
+ * Only removes the last line if its "episode_index" field matches the
+ * expected index. Returns true (no-op) if the last entry doesn't match,
+ * preventing accidental deletion of a previous episode's metadata.
  *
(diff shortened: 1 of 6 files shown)


shantanuparab-tr:
* **#206** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/206?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#205** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/205?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#204** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/204?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/204?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#203** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/203?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#202** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/202?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#201** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/201?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

include/trossen_sdk/io/backends/lerobot_v2/lerobot_v2_backend.hpp
```diff
@@ -417,6 +417,92 @@ inline bool write_task_entry(
...
+ * Used during episode discard to undo appended entries.
+ *
+ * @param file_path Path to the JSONL file
+ * @return true on success, false on failure
+ */
+inline bool remove_last_jsonl_line(const std::filesystem::path& file_path) {
+  namespace fs = std::filesystem;
+  if (!fs::exists(file_path)) return true;
+
+  std::ifstream in(file_path);
```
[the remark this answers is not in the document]
shantanuparab-tr: Will be adding tests later.  We don't directly collect epsidoes in LeRobot format and therefore there is no way of actually testing these changes. We have implemented them here as they make more logical sense in this PR . Also, discard_episode is a virtual method and would break the build if not implemented.

include/trossen_sdk/io/backends/lerobot_v2/lerobot_v2_backend.hpp
```diff
@@ -417,6 +417,92 @@ inline bool write_task_entry(
...
+
+  if (lines.empty()) return true;
+  lines.pop_back();
+
+  std::ofstream out(file_path, std::ios::trunc);
+  if (!out.is_open()) return false;
+  for (const auto& l : lines) {
+    out << l << "\n";
+  }
+  out.close();
```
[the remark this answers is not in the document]
shantanuparab-tr: We write one episode at a time so the list is not going to grow too big. In future, we can handle this better if needed.

include/trossen_sdk/io/backends/lerobot_v2/lerobot_v2_backend.hpp
```diff
@@ -417,6 +417,92 @@ inline bool write_task_entry(
...
+  info_json["total_chunks"] = std::max(1, (total_episodes + chunks_size - 1) / chunks_size);
+
+  // Adjust train split end
+  std::string train_split = info_json["splits"].value("train", "0:0");
+  size_t colon_pos = train_split.find(':');
+  if (colon_pos != std::string::npos) {
+    int train_start = std::stoi(train_split.substr(0, colon_pos));
+    int train_end = std::max(train_start, std::stoi(train_split.substr(colon_pos + 1)) - 1);
+    info_json["splits"]["train"] = std::to_string(train_start) + ":" + std::to_string(train_end);
+  }
```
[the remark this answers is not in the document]
shantanuparab-tr: We add missing key guard along with garbage value check. Additionally, we also wrap it in a try catch to protect against any edge cases.

src/backends/lerobot_v2/lerobot_v2_backend.cpp
```diff
@@ -828,28 +824,134 @@ void LeRobotV2Backend::close() {
...
+void LeRobotV2Backend::close() {
+  std::lock_guard<std::mutex> lock(open_mutex_);
+
+  close_resources();
 
   // Encode any remaining images to videos
   convert_to_videos();
 
   // Compute dataset statistics
   compute_statistics();
```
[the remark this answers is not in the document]
shantanuparab-tr: added early return in close when backend was never opened.

src/backends/lerobot_v2/lerobot_v2_backend.cpp
```diff
@@ -802,14 +802,11 @@ void LeRobotV2Backend::flush() {
...
 }
 
-void LeRobotV2Backend::close() {
-  std::lock_guard<std::mutex> lock(open_mutex_);
-  if (!opened_) {
-    return;
-  }
+void LeRobotV2Backend::close_resources() {
+  // Caller must hold open_mutex_. Closes writer, stream, and image workers.
+  if (!opened_) return;
```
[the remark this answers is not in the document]
shantanuparab-tr: Now close_resources() only bails if there's truly nothing to clean up. No open flag, no
writer, no output stream, and no threads. If open() threw after starting threads but  
before setting opened\_, the image_workers_.empty() check fails and cleanup proceeds normally.

include/trossen_sdk/io/backends/lerobot_v2/lerobot_v2_backend.hpp
```diff
@@ -417,6 +417,116 @@ inline bool write_task_entry(
...
+  std::vector<std::string> lines;
+  std::string line;
+  while (std::getline(in, line)) {
+    if (!line.empty()) lines.push_back(line);
+  }
+  in.close();
+
+  if (lines.empty()) return true;
+
+  // Only remove if the last entry's episode_index matches
```
lukeschmitt-tr, line 449: When might this happen? Might this introduce any unexpected behavior?  
  
This will "fail" silently, but data is not corrupted

src/backends/lerobot_v2/lerobot_v2_backend.cpp
```diff
@@ -802,14 +802,13 @@ void LeRobotV2Backend::flush() {
...
-void LeRobotV2Backend::close() {
-  std::lock_guard<std::mutex> lock(open_mutex_);
-  if (!opened_) {
-    return;
-  }
+void LeRobotV2Backend::close_resources() {
+  // Caller must hold open_mutex_. Closes writer, stream, and image workers.
+  // Check actual resource state, not just opened_, because open() may have
+  // started threads before a later step threw (leaving opened_ == false).
+  if (!opened_ && !writer_ && !outfile_ && image_workers_.empty()) return;
```
lukeschmitt-tr, line 809: why might this fail, and does a silent return cause us any issues?

src/backends/lerobot_v2/lerobot_v2_backend.cpp
```diff
@@ -828,28 +826,137 @@ void LeRobotV2Backend::close() {
...
+      }
+      if (!last_line.empty()) {
+        auto entry = nlohmann::json::parse(last_line);
+        if (entry.value("episode_index", -1) == static_cast<int>(episode_index_)) {
+          discarded_frame_count = entry.value("length", 0);
+        }
+      }
+    }
+  } catch (...) {
+    // Best-effort: if we can't read the frame count, counters will be approximate
```
lukeschmitt-tr, line 932: What effect does this have? Can we warn?

lukeschmitt-tr (changes requested):

src/backends/lerobot_v2/lerobot_v2_backend.cpp
shantanuparab-tr: silent return shouldn't be a problem as all the resources are already closed and we move to discarding the episode. Didn't get a chance to verify why it might fail as we don't collect using the LeRobot framework.  
This covers most of the cases not all.

src/backends/lerobot_v2/lerobot_v2_backend.cpp
shantanuparab-tr: We can warn, this code isn't fully tested so wasn't able to capture any effect. I only predicted what could happen and implemented this accordingly. The plan was to get the scaffolding ready, so that we don't miss out on things when we actually implement the LeRobot backend for data capture. These new methods do not affect the conversion script as of now.

include/trossen_sdk/io/backends/lerobot_v2/lerobot_v2_backend.hpp
shantanuparab-tr: The episode json is updated at the end of the episode. However, discard episode can be called while the episode is getting recorded. If we delete the json line without checking the episode index we might delete previous episodes data.  
  
If line is not found it can fail silently as there was nothing to delete.

lukeschmitt-tr (approved):

shantanuparab-tr:
### Merge activity

* **Apr 2, 2:46 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/204).
* **Apr 2, 2:51 PM UTC**: [Graphite](https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/204) rebased this pull request as part of a merge.
* **Apr 2, 2:52 PM UTC**: @shantanuparab-tr merged this pull request with [Graphite](https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/204).

PR #216, opened 2026-04-13
backends: parallelize mcap to lerobot v2 conversion

alibabaistaken:
## Summary
Converts the MCAP to LeRobot v2 converter from a sequential single-threaded
script into a fully parallel GPU-accelerated pipeline.

### GPU auto-detection
Detects the best available video encoder at startup and uses it automatically:
- NVIDIA GPU → av1_nvenc
- AMD GPU    → av1_amf
- Intel GPU  → av1_qsv
- CPU        → libsvtav1 (fallback, always works)
- No encoder → error + exit

User can override via `--encoder gpu|cpu|auto` flag (CLI > config > auto).
Per-camera GPU fallback: if NVENC rejects a session (exit 47872), that
camera automatically retries with libsvtav1. No episode ever fails due
to GPU unavailability.

### Episode parallelism
Multiple episodes convert simultaneously via a true thread pool.
Exactly num_threads persistent workers (default 70% of CPU cores, clamped
to hw_threads max) pull from a shared mutex-protected task queue — no
condition variable or done flag needed since tasks are fully populated
before workers spawn. 500 episodes creates 16 threads total, not 500.

### Parallel image extraction
Within each episode, image extraction uses a producer-consumer pattern:
1 producer reads the MCAP sequentially, 8 per-camera worker threads
decode and write JPGs simultaneously. Encoding only starts after all
camera workers finish.

### Video encoding
All 8 cameras within one episode encode simultaneously. A global
ep_encode_sem (count=1) serializes which episode encodes at any moment,
preventing NVENC session overflow (laptop GPUs allow ~3-5 concurrent
sessions). While one episode encodes, other episodes freely run MCAP
read and image extraction phases — no idle time.

### FFmpeg via posix_spawnp
Replaced all `std::system()` ffmpeg calls with `posix_spawnp`. FFmpeg
now receives arguments directly as an argv array with no shell involved,
so paths with spaces or special characters work correctly without any
quoting or escaping. stderr is redirected to `/dev/null` via
`posix_spawn_file_actions_addopen`.

### User controls
- `--threads N`    → override worker count. Clamped to hw_threads with
                     warning when exceeded.
- `--encoder gpu`  → require GPU encoder, error if none found
- `--encoder cpu`  → skip GPU probe, use libsvtav1 directly
- `--encoder auto` → auto-detect (default)

Output shows chosen encoder and thread count with source:
  Video encoder: NVIDIA GPU (av1_nvenc)  [auto-detected]
  Using 16 core(s) of 24 available  (auto: 70% of 24)

### Global index contiguity
Each episode is pre-assigned a contiguous starting index before workers
spawn. A lightweight MCAP footer scan (readSummary) estimates row count
per episode, cumulative offsets are computed in episode order, and each
worker uses a local counter starting at its assigned offset. This
guarantees the global index column is contiguous per episode even when
episodes run in parallel — required by the LeRobot v2 validator.

### Metadata thread safety
All writes to shared meta/ files (info.json, tasks.jsonl, episodes.jsonl,
episodes_stats.jsonl) are serialized under a shared meta_mutex. Without
this, concurrent episodes could corrupt JSONL files via interleaved
appends or lose info.json updates via read/modify/write races.

### Thread safety
- Episode output buffered into per-episode ostringstream, printed
  atomically under cout_mutex on completion — no interleaved output.
- All chunk directories pre-created in main() before workers spawn
  using std::error_code overload to eliminate directory race conditions.
- Worker lambda wrapped in try/catch so exceptions never leave the pool
  in a broken state.
- ep_encode_sem wrapped in RAII guard — semaphore always released on
  every exit path including exceptions.
- global_index_counter initialized from existing parquet max index on
  resume so re-runs never produce duplicate row indices.

### Skip/resume
On re-run after crash, each episode checks if its parquet exists and is
valid (opens + reads row count). Valid → skip. Corrupt/missing → delete
and re-convert. Safe to resume from last good episode.

### Per-stage timing
Each episode logs a breakdown on completion:
  MCAP read / Parquet write / Image extraction / Video encoding / Metadata / Total

## Performance (5 episodes, Trossen Stationary AI, 8 cameras/episode)

| Version | Wall clock | Video/ep | Image extract/ep | Note |
|---|---|---|---|---|
| Original (sequential) | 74.4s | ~40s | N/A | No parallelism, CPU only |
| Parallel CPU | 52.4s | ~38s | ~2.6s | Episode parallel, libsvtav1 |
| Parallel GPU (working) | 35.7s | ~3.3s | ~2s | GPU wired correctly |
| After initial PR fixes | 23s | ~13s | ~2.3s | True thread pool, all fixes |

Video encoding per episode is higher than the early GPU result (~3.3s)
because 16 concurrent episodes compete for the laptop GPU's NVENC session
limit (~3-5 sessions). The semaphore prevents overflow — cameras that
cannot acquire a session fall back to CPU automatically.

Scale test: 199 episodes → 15m49s wall clock (4.76s/ep).
Cross-machine: RTX 5090 Desktop → 10.45s for 5 episodes (2.09s/ep).

## Validation
Dataset validated against the Trossen LeRobot edit_dataset.py validator:
all 8 checks pass including parquet contiguity, video file presence,
task references, and stats consistency.

## Tested on
- Machine: 24-core CPU, NVIDIA RTX 5070 Ti Laptop GPU
- Dataset: Trossen Stationary AI (5 episodes, 8 cameras)
- Scale test: 199 episodes → 15m49s (4.76s wall clock/ep)
- Cross-machine: RTX 5090 Desktop → 10.45s for 5 episodes (2.09s/ep)

Relates to TDS-98

### DIFF
--- a/scripts/trossen_mcap_to_lerobot_v2/trossen_mcap_to_lerobot_v2.cpp
+++ b/scripts/trossen_mcap_to_lerobot_v2/trossen_mcap_to_lerobot_v2.cpp
@@ -230,7 +230,9 @@ nlohmann::ordered_json compute_episode_stats(const std::filesystem::path& parque
 int process_mcap_file(const std::string& mcap_file, const std::string& dataset_root_dir,
                       int episode_index, const std::string& repository_id,
                       const std::string& dataset_id, int chunk_size,
-                      const std::string& video_encoder, const std::string& encoder_flags);
+                      const std::string& video_encoder, const std::string& encoder_flags,
+                      std::atomic<int64_t>& global_index_counter,
+                      std::ostringstream& log);
 
 // ──────────────────────────────────────────────────────────
 // Main entry point
@@ -368,8 +370,8 @@ int main(int argc, char** argv) {
   if (hw_threads == 0) hw_threads = 4;
   unsigned int default_threads = std::max(1u, static_cast<unsigned int>(hw_threads * 0.7));
   int requested_threads = cli.get_int("threads", static_cast<int>(default_threads));
-requested_threads = std::max(1, requested_threads);
-unsigned int num_threads = static_cast<unsigned int>(requested_threads);
+  requested_threads = std::max(1, std::min(requested_threads, static_cast<int>(hw_threads)));
+  unsigned int num_threads = static_cast<unsigned int>(requested_threads);
   std::cout << "Using " << num_threads << " core(s) of " << hw_threads << " available\n\n";
 
   // Detect best available video encoder
@@ -453,61 +455,100 @@ unsigned int num_threads = static_cast<unsigned int>(requested_threads);
     tasks.push_back({mcap_path, episode_index, i + 1});
   }
 
+  // Pre-create all chunk directories before spawning workers to avoid
+  // concurrent fs::create_directories calls on overlapping paths.
+  for (const auto& task : tasks) {
+    int ep_chunk = task.episode_index / chunk_size;
+    fs::path chunk_data = fs::path(dataset_root_dir) / repository_id / dataset_id / "data" /
+                          trossen::io::backends::format_chunk_dir(ep_chunk);
+    fs::path chunk_images = fs::path(dataset_root_dir) / repository_id / dataset_id / "images" /
+                            trossen::io::backends::format_chunk_dir(ep_chunk);
+    fs::path chunk_videos = fs::path(dataset_root_dir) / repository_id / dataset_id / "videos" /
+                            trossen::io::backends::format_chunk_dir(ep_chunk);
+    fs::path meta_dir = fs::path(dataset_root_dir) / repository_id / dataset_id / "meta";
+    std::error_code ec;
+    fs::create_directories(chunk_data, ec);
+    fs::create_directories(chunk_images, ec);
+    fs::create_directories(chunk_videos, ec);
+    fs::create_directories(meta_dir, ec);
+  }
+  std::atomic<int64_t> global_index_counter{0};
   auto total_start = std::chrono::steady_clock::now();
 
   if (!tasks.empty()) {
-    // Dispatch episodes in parallel, capped at num_threads concurrent cores
-    std::mutex pool_mutex;
-    std::condition_variable pool_cv;
-    unsigned int active = 0;
     std::mutex cout_mutex;
+    std::mutex queue_mutex;
+    std::condition_variable queue_cv;
+    size_t task_index = 0;
+    bool done = false;
 
-    std::vector<std::thread> threads;
-    threads.reserve(tasks.size());
-
-    for (const auto& task : tasks) {
-      // Block until a core is free
-      {
-        std::unique_lock<std::mutex> lock(pool_mutex);
-        pool_cv.wait(lock, [&] { return active < num_threads; });
-        ++active;
-      }
-
-      threads.emplace_back([&, task]() {
-        auto ep_start = std::chrono::steady_clock::now();
-        int result;
+    auto worker = [&]() {
+      while (true) {
+        EpisodeTask task;
         {
-          std::lock_guard<std::mutex> lk(cout_mutex);
-          std::cout << "\n" << std::string(70, '=') << "\n";
-          std::cout << "Processing file " << task.display_idx << "/" << mcap_files.size() << ": "
-                    << task.mcap_path.filename().string() << "\n";
-          std::cout << std::string(70, '=') << "\n";
+          std::unique_lock<std::mutex> lock(queue_mutex);
+          queue_cv.wait(lock, [&] { return task_index < tasks.size() || done; });
+          if (task_index >= tasks.size()) return;
+          task = tasks[task_index++];
+        }
+        auto ep_start = std::chrono::steady_clock::now();
+        int result = 1;
+        try {
+          {
+            std::lock_guard<std::mutex> lk(cout_mutex);
+            std::cout << "\n" << std::string(70, '=') << "\n";
+            std::cout << "Processing file " << task.display_idx << "/" << mcap_files.size()
+                      << ": " << task.mcap_path.filename().string() << "\n";
+            std::cout << std::string(70, '=') << "\n";
+          }
+          std::ostringstream ep_log;
           result = process_mcap_file(task.mcap_path.string(), dataset_root_dir,
                                      task.episode_index, repository_id, dataset_id, chunk_size,
-                                     video_encoder, encoder_flags);
-          auto ep_end = std::chrono::steady_clock::now();
-          double ep_secs = std::chrono::duration<double>(ep_end - ep_start).count();
-          std::cout << "[TIMING] Episode " << task.episode_index << " took " << std::fixed
-                    << std::setprecision(2) << ep_secs << "s\n";
-          if (result != 0) {
-            std::cerr << "\n[FAILED] Failed to process: " << task.mcap_path.string() << "\n";
+                                     video_encoder, encoder_flags,
+                                     global_index_counter, ep_log);
+          {
+            std::lock_guard<std::mutex> lk(cout_mutex);
+            std::cout << ep_log.str();
+            auto ep_end = std::chrono::steady_clock::now();
+            double ep_secs = std::chrono::duration<double>(ep_end - ep_start).count();
+            std::cout << "[TIMING] Episode " << task.episode_index << " took " << std::fixed
+                      << std::setprecision(2) << ep_secs << "s\n";
+            if (result != 0) {
+              std::cerr << "\n[FAILED] Failed to process: " << task.mcap_path.string() << "\n";
+            }
           }
+        } catch (const std::exception& e) {
+          std::lock_guard<std::mutex> lk(cout_mutex);
+          std::cerr << "\n[EXCEPTION] Episode " << task.episode_index
+                    << " threw: " << e.what() << "\n";
+          result = 1;
+        } catch (...) {
+          std::lock_guard<std::mutex> lk(cout_mutex);
+          std::cerr << "\n[EXCEPTION] Episode " << task.episode_index
+                    << " threw unknown exception\n";
+          result = 1;
         }
         if (result == 0) {
           ++successful;
         } else {
           ++failed;
         }
-        {
-          std::lock_guard<std::mutex> lock(pool_mutex);
-          --active;
-        }
-        pool_cv.notify_one();
-      });
-    }
+      }
+    };
 
-    for (auto& t : threads) {
-      t.join();
+    std::vector<std::thread> workers;
+    workers.reserve(num_threads);
+    for (unsigned int i = 0; i < num_threads; ++i) {
+      workers.emplace_back(worker);
+    }
+    queue_cv.notify_all();
+    {
+      std::lock_guard<std::mutex> lock(queue_mutex);
+      done = true;
+    }
+    queue_cv.notify_all();
+    for (auto& w : workers) {
+      w.join();
     }
   }
 
@@ -634,7 +675,9 @@ unsigned int num_threads = static_cast<unsigned int>(requested_threads);
 int process_mcap_file(const std::string& mcap_file, const std::string& dataset_root_dir,
                       int episode_index, const std::string& repository_id,
                       const std::string& dataset_id, int chunk_size,
-                      const std::string& video_encoder, const std::string& encoder_flags) {
+                      const std::string& video_encoder, const std::string& encoder_flags,
+                      std::atomic<int64_t>& global_index_counter,
+                      std::ostringstream& log) {
   ParquetConfig cfg;
   cfg.mcap_file = mcap_file;
   cfg.dataset_root = dataset_root_dir;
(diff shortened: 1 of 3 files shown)


scripts/trossen_mcap_to_lerobot_v2/trossen_mcap_to_lerobot_v2.cpp
```diff
@@ -413,23 +450,75 @@ int main(int argc, char** argv) {
...
+          std::lock_guard<std::mutex> lk(cout_mutex);
+          std::cout << "\n" << std::string(70, '=') << "\n";
+          std::cout << "Processing file " << task.display_idx << "/" << mcap_files.size() << ": "
+                    << task.mcap_path.filename().string() << "\n";
+          std::cout << std::string(70, '=') << "\n";
+          result = process_mcap_file(task.mcap_path.string(), dataset_root_dir,
+                                     task.episode_index, repository_id, dataset_id, chunk_size,
+                                     video_encoder, encoder_flags);
+          auto ep_end = std::chrono::steady_clock::now();
+          double ep_secs = std::chrono::duration<double>(ep_end - ep_start).count();
```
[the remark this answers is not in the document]
alibabaistaken: > The episode worker thread holds `cout_mutex` while calling `process_mcap_file(...)`, which serializes the entire conversion and effectively defeats the intended episode-level parallelism. Limit the `cout_mutex` scope to just the banner/timing prints (or buffer per-episode logs) and call `process_mcap_file` outside the locked section so multiple episodes can run concurrently.

The cout_mutex intentionally wraps process_mcap_file to prevent garbled output 
from concurrent episodes printing simultaneously. Moving it outside would restore 
true parallelism and improve wall clock time, but at the cost of interleaved 
terminal output from multiple episodes printing at the same time.

scripts/trossen_mcap_to_lerobot_v2/trossen_mcap_to_lerobot_v2.cpp
```diff
@@ -1393,34 +1494,44 @@ int process_mcap_file(const std::string& mcap_file, const std::string& dataset_r
...
-        auto duration =
-            std::chrono::duration_cast<std::chrono::milliseconds>(encode_end - encode_start)
-                .count();
-        std::cout << " [ok] (" << (duration / 1000.0) << "s)\n";
-        videos_created++;
-      } else {
-        std::cout << " [FAILED] Failed (exit code " << ret << ")\n";
-        std::cerr << "    Command: " << ffmpeg_cmd.str() << "\n";
+                 << " -c:v " << video_encoder << " " << encoder_flags << " -pix_fmt yuv420p -r 30 "
+                 << video_output.string() << " 2>/dev/null";
```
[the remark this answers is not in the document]
alibabaistaken: > The ffmpeg invocation is built as a single shell command and passed to `std::system` with unquoted/unsanitized paths (`input_pattern`/`video_output`) and `2>/dev/null` redirection. If any directory or camera name contains spaces or shell metacharacters, encoding can fail or (worse) allow command injection. Prefer spawning ffmpeg without a shell (e.g., `posix_spawnp` with an argv array like in `trossen::utils::announce`), or at minimum robustly quote/escape all interpolated arguments and avoid shell redirection.

All paths come from our own config and MCAP files, not user input, so 
injection isn't a real risk here.

scripts/trossen_mcap_to_lerobot_v2/trossen_mcap_to_lerobot_v2.cpp
```diff
@@ -588,7 +926,7 @@ int process_mcap_file(const std::string& mcap_file, const std::string& dataset_r
 
   trossen::utils::print_config_banner("TrossenMCAP to LeRobotV2 Converter", config_lines);
```
shantanuparab-tr: The per-episode log buffering is a nice fix for interleaving, but print_config_banner here writes straight to std::cout (see src/utils/app_utils.cpp:164). Under parallel episodes those banners will still interleave with each other and with the flushed ep_log. Either add a stream overload for print_config_banner and route it through log, or just build the banner lines into log directly like the rest of the function does.

scripts/trossen_mcap_to_lerobot_v2/trossen_mcap_to_lerobot_v2.cpp
```diff
@@ -367,39 +407,206 @@ int main(int argc, char** argv) {
...
+    // (throws on open), delete it and re-add to the task list for re-conversion.
+    // This makes re-runs safe — kill the script and restart without losing progress.
     if (fs::exists(expected_parquet)) {
-      // Advance global index past the skipped episode's rows.
-      // If the parquet is corrupt / unreadable, delete it and fall through
-      // to re-convert so we don't end up with overlapping indices.
       try {
         auto reader = parquet::ParquetFileReader::OpenFile(expected_parquet.string(), false);
-        global_index_offset += reader->metadata()->num_rows();
+        reader->metadata()->num_rows();  // throws if file is corrupt  // throws if file is corrupt
```
shantanuparab-tr: Small one: the inline comment got duplicated (`// throws if file is corrupt  // throws if file is corrupt`). Also worth saying explicitly that the return value is intentionally discarded — this is a liveness probe, not dead code — so future readers don't try to clean it up.

scripts/trossen_mcap_to_lerobot_v2/trossen_mcap_to_lerobot_v2.cpp
```diff
@@ -601,48 +939,41 @@ int process_mcap_file(const std::string& mcap_file, const std::string& dataset_r
     fs::create_directories(videos_dir);
     fs::create_directories(meta_dir);
```
shantanuparab-tr: Main already pre-creates the chunk dirs at lines 629-643 before spawning workers (good fix for the race). These calls inside process_mcap_file are now redundant and use the throwing overload, so they're no-ops in the happy path but can still throw on unrelated FS errors and bubble up into the worker. Can we just drop them?

scripts/trossen_mcap_to_lerobot_v2/trossen_mcap_to_lerobot_v2.cpp
```diff
@@ -367,39 +407,206 @@ int main(int argc, char** argv) {
...
+  int config_threads = lerobot_config->threads;
+  unsigned int num_threads;
+  std::string threads_source;
+
+  if (cli_threads > 0) {
+    if (static_cast<unsigned int>(cli_threads) > hw_threads) {
+      std::cout << "[WARNING] --threads " << cli_threads
+                << " exceeds hardware concurrency (" << hw_threads
+                << "). Falling back to auto (" << default_threads << ").\n";
+      num_threads = default_threads;
```
shantanuparab-tr: When the user passes --threads 50 on a 24-core machine, we warn and then silently drop their value and use default_threads (70% of hw_threads). That throws away explicit intent. Intuitively I'd expect a clamp-to-hw_threads instead. If the fallback is intentional please add a note, and either way consider echoing the value you actually ended up using so the user sees what happened. Same thing applies to the config-based branch a few lines down.

scripts/trossen_mcap_to_lerobot_v2/trossen_mcap_to_lerobot_v2.cpp
```diff
@@ -1124,36 +1452,37 @@ int process_mcap_file(const std::string& mcap_file, const std::string& dataset_r
...
-        std::cerr << "Error: Failed to append action value\n";
+        log << "Error: Failed to append action value\n";
         return 1;
       }
     }
 
     if (!epi_idx_builder.Append(cfg.episode_index).ok() ||
-        !frame_idx_builder.Append(frame_index).ok() || !index_builder.Append(global_index).ok() ||
+        !frame_idx_builder.Append(frame_index).ok() ||
+        !index_builder.Append(global_index_counter.fetch_add(1)).ok() ||
```
shantanuparab-tr, line 1626: Switching to fetch_add guarantees globally unique indices, which is what #213 was after. But with parallel episodes the rows of a single episode are no longer contiguous in the index column (episode 0 might get [0, 4, 9, ...] if episode 1 is writing concurrently). HF's key-lookup path is fine with that, but anything that slices by index range expecting "episode N = indices [a..b]" will break. Worth sanity-checking against a downstream loader and calling it out in the PR description.

scripts/trossen_mcap_to_lerobot_v2/trossen_mcap_to_lerobot_v2.cpp
```diff
@@ -413,24 +620,141 @@ int main(int argc, char** argv) {
...
+    // spawn exactly num_threads persistent workers — not one per episode
+    std::vector<std::thread> workers;
+    workers.reserve(num_threads);
+    for (unsigned int i = 0; i < num_threads; ++i) {
+      workers.emplace_back(worker);
+    }
+    queue_cv.notify_all();  // wake all workers so they start pulling tasks immediately
+    {
+      std::lock_guard<std::mutex> lock(queue_mutex);
+      done = true;  // signal that no more tasks will be added
```
shantanuparab-tr: Since tasks is fully populated before workers spawn and done is set to true immediately after spawning, the CV-wait never actually needs to block — workers always find either more work or an exhausted queue. You could simplify to a plain while-loop that takes the mutex, checks task_index < tasks.size(), pops, and returns on exhaustion — no done flag, no CV. Not blocking, just noting the machinery isn't pulling its weight.

scripts/trossen_mcap_to_lerobot_v2/trossen_mcap_to_lerobot_v2.cpp
```diff
@@ -327,8 +362,12 @@ int main(int argc, char** argv) {
   const std::string dataset_id = lerobot_config->dataset_id;
   const int chunk_size = lerobot_config->chunk_size;
   const std::string license = lerobot_config->license;
+  const std::string config_task_name = lerobot_config->task_name;
+  const std::string config_robot_name = lerobot_config->robot_name;
+  const float config_fps = lerobot_config->fps;
+  const bool overwrite_existing = lerobot_config->overwrite_existing;
+  const bool encode_videos = lerobot_config->encode_videos;
```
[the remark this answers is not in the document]
alibabaistaken: added!

scripts/trossen_mcap_to_lerobot_v2/trossen_mcap_to_lerobot_v2.cpp
```diff
@@ -367,39 +411,206 @@ int main(int argc, char** argv) {
...
+
+  // Thread count resolution — priority: CLI > config > auto 70%
+  // If the requested value exceeds hardware concurrency, warn and fall back to auto.
+  unsigned int default_threads = std::max(1u, static_cast<unsigned int>(hw_threads * 0.7));
+  int cli_threads = cli.get_int("threads", -1);
+  int config_threads = lerobot_config->threads;
+  unsigned int num_threads;
+  std::string threads_source;
+
+  if (cli_threads > 0) {
```
shantanuparab-tr, line 433: we have 2 different behviour based on where the configuration is read from for cli_threads we calm to default_thread and for config_thread we use hw_threads

scripts/trossen_mcap_to_lerobot_v2/trossen_mcap_to_lerobot_v2.cpp
```diff
@@ -1393,49 +1917,91 @@ int process_mcap_file(const std::string& mcap_file, const std::string& dataset_r
       fs::path input_pattern = camera_dir / "image_%06d.jpg";
 
       std::ostringstream ffmpeg_cmd;
-      // Force output to exactly 30fps for perfect timestamp alignment
-      ffmpeg_cmd << "ffmpeg -y -loglevel error -framerate " << cfg.camera_fps << " -start_number 0"
-                 << " -i " << input_pattern.string() << " -frames:v "
+      ffmpeg_cmd << "ffmpeg -y -loglevel error -framerate " << cfg.camera_fps
```
shantanuparab-tr: we should use posix_spawnp("ffmpeg", argv...) to avoid using the shell. This will help avoid shell command corruption because of escape characters.

scripts/trossen_mcap_to_lerobot_v2/trossen_mcap_to_lerobot_v2.cpp
shantanuparab-tr, line 1: We should update the readme file for this example to cover the updates

shantanuparab-tr (changes requested):

scripts/trossen_mcap_to_lerobot_v2/trossen_mcap_to_lerobot_v2.cpp
alibabaistaken: > we should use posix_spawnp("ffmpeg", argv...) to avoid using the shell. This will help avoid shell command corruption because of escape characters.

 replaced std::system() with posix_spawnp. FFmpeg now receives arguments directly as an argv array with no shell involved

scripts/trossen_mcap_to_lerobot_v2/trossen_mcap_to_lerobot_v2.cpp
alibabaistaken: done

PR #241, opened 2026-05-18
observers: integrate with SessionManager via producer fan-out and lazy lifecycle

shantanuparab-tr:
## Summary

Wire observers into `SessionManager`. Producers now fan out each record to `sink.enqueue `plus `observer.offer × N` inline in their emit callback. Observers start lazily on the first `start_episode()`, persist across episode boundaries, and are stopped at `shutdown()` (ADR-003 L2 lifecycle). A failed `start()` marks the observer dead and the session continues without it.

## Changes

- Add `SessionManager::add_observer()` to register an observer before the first episode.
- Add producer-side fan-out: a single `fan_out_record_()` helper used by both polled and push producer emit lambdas, with a no-observer fast path that is one branch over the pre-existing sink-only flow.
- Lazy-start observers at first `start_episode()`; snapshot the live (non-dead) observers into the producer lambdas for the episode.
- Stop observers in `shutdown()` after recording has stopped, and refuse `start_episode()` once shutdown has been called.
- Add `SessionManager::observer_stats()` returning per-observer counters + `is_running` / `is_dead` so example apps can surface observer health.
- Add `tests/test_session_manager_observers.cpp` covering: null-observer rejection, late-add rejection, dead-observer exclusion from fan-out, fan-out reaches every observer, observer survives episode boundaries, shutdown gates further episodes.

## Test Plan

- [x] Builds cleanly (`make build`)
- [x] Tests pass (`make test`) — new `test_session_manager_observers` suite plus the existing session-manager suites
- [x] Lint passes (`pre-commit run --all-files`)
- [x] Tested on hardware — pending; covered by the example wiring in PR 6

## Breaking Changes

None. All additions are new public methods; existing `SessionManager` callers keep working without changes.

## Related Issues

Relates to TDS-116

Relates to TDS-170

PR #240, opened 2026-05-18
observers: add ObserverConfig schema and SdkConfig observer validation

shantanuparab-tr:
## Summary

Add `ObserverConfig` JSON schema and wire it into the top-level `SdkConfig` so observers can be declared in the unified robot config alongside producers and backends. Validates duplicate observer ids and warns when a subscription's `record_id` doesn't match any
producer's `stream_id` (catches typos that would otherwise silently no-op).

## Changes

- Add `include/trossen_sdk/configuration/types/observers/observer_config.hpp` with `ObserverConfig` + `ObserverSubscriptionConfig`. Common fields parsed; raw JSON preserved so registry factories can read transport-specific fields.
- Update `SdkConfig::from_json()` to parse the optional `"observers"` array, reject duplicate ids, and warn on subscription `record_id`s that no producer emits.

## Test Plan

- [x] Builds cleanly (`make build`)
- [x] Tests pass (`make test`)
- [x] Lint passes (`pre-commit run --all-files`)
- [x] Tested on hardware (N/A — config parsing only)

## Breaking Changes

None — the `observers` JSON key is optional; configs without it continue to load unchanged.

## Related Issues

Relates to TDS-116

Relates to TDS-172

PR #242, opened 2026-05-18
observers: add RerunObserver with optional rerun-cpp FetchContent build gate

shantanuparab-tr:
## Summary

Add `RerunObserver`, the first concrete observer: streams records to a live ReRun viewer over gRPC. Gated behind a CMake option (`TROSSEN_ENABLE_RERUN_OBSERVER=OFF` by default) because rerun-cpp pulls ~50 MB and adds multiple minutes to first build — only the
stationary demo wants it today.

## Changes

- Add `include/trossen_sdk/observer/rerun_observer.hpp` (forward-declares `rerun::RecordingStream` so rerun headers never leak into the public include surface).
- Add `src/observer/rerun_observer.cpp` with the dispatch: `ImageRecord` → `rerun::Image `(+ optional depth), `JointStateRecord` → three `rerun::Scalars` paths (positions / velocities / efforts). Counts and logs unsupported encodings / record types via `skipped_frames()` so a silent-empty viewer is debuggable.
- Add `TROSSEN_ENABLE_RERUN_OBSERVER` option in top-level `CMakeLists.txt` with FetchContent for rerun-cpp v0.32.0 (pinned SHA256), reuses the system Arrow that trossen_sdk already links, sets PIC for the SHARED build, and adds the `--allow-multiple-definition` linker workaround for the duplicate Rust runtime symbols that ship with `librerun_c` + `libfoxglove`.
- Add an explicit `Arrow >= 14.0` floor gated on `TROSSEN_ENABLE_RERUN_OBSERVER`. rerun-cpp 0.32.0 reuses the host Arrow via `RERUN_ARROW_LINK_SHARED=ON`; an ABI-incompatible Arrow links silently and crashes at runtime instead of failing at build time. The check fires only when the rerun observer is enabled, so default builds keep the same Arrow surface they always had.
- Add `tests/test_rerun_observer.cpp` (gated on the same CMake option) covering config defaults, missing-fields rejection, and the `mat_to_rgb_bytes` encoding helper.

## Symbol-collision audit (--allow-multiple-definition)

`librerun_c.a` and `libfoxglove.a` each bundle the Rust standard library, so the linker
sees duplicate Rust runtime symbols. `--allow-multiple-definition` lets it keep one
copy. Audit run on Ubuntu 24.04 / clang-18 / Arrow 24.0.0 / rerun-cpp 0.32.0 /
foxglove 0.16.1:

```
$ nm -A --defined-only build/libtrossen_sdk.so | awk '{print $3}' | sort | uniq -d | wc -l
1208
```

Categorized breakdown of the 1208 duplicates:

| Count | Category | Notes |
|------:|----------|-------|
|   334 | Rust stdlib symbols (`_$LT$std..sys$GT$...`, `core::...`, `alloc::...`) | The expected duplication between librerun_c and libfoxglove; both static-link the Rust standard library at the same major version. |
|   418 | `GCC_except_table*` entries | Per-function exception unwind metadata in readonly file-local data. The linker keeps the relocations pointing at the right table; duplicate names are bookkeeping. |
|    21 | zstd compression internals (`FSE_*`, `HUF_*`, `BIT_*`, `algoTime`, …) | zstd is statically linked from Arrow / Parquet / Foxglove / MCAP; the duplicates are identical read-only lookup tables. |
|    10 | C++ template / instantiation weak symbols (`std::function`, `std::map` specialisations) | Normal C++ weak-linkage merging; would be folded by the linker regardless of this flag. |
|  ~425 | `_GLOBAL__I_*` static-init wrappers, anonymous-namespace helpers, file-local readonly data | Linker bookkeeping. |

**Zero trossen_sdk public-API symbols collide.** Verified with:

```
$ grep '^_ZN7trossen[0-9]\+[a-z]' /tmp/dup-syms.txt | grep -v 'configuration[0-9]*\$_'
(empty)
```

The flag is therefore safe for the documented scope: it lets the Rust-runtime
collisions resolve without spuriously breaking the link, and every duplicate category
above tolerates the linker keeping one copy. The flag is gated behind
`TROSSEN_ENABLE_RERUN_OBSERVER` (default OFF), so default builds never see it.

## Test Plan

- [x] Builds cleanly (`make build`) with the option OFF
- [x] Builds cleanly with `-DTROSSEN_ENABLE_RERUN_OBSERVER=ON`
- [x] Arrow floor check fires when Arrow < 14.0 (verified by CMake conditional; passes against Arrow 24.0.0 on the dev host)
- [x] Symbol-collision audit attached above
- [x] Tests pass (`make test`) — `test_rerun_observer` runs only when the option is ON
- [x] Lint passes (`pre-commit run --all-files`)
- [x] Tested on hardware — connecting to a running `rerun --serve` viewer and confirming frames appear; covered end-to-end in PR 6

## Breaking Changes

None. The CMake option defaults OFF, so default builds are byte-identical to before. The new Arrow floor check only fires when the option is ON.

## Related Issues

Relates to TDS-116

Relates to TDS-173

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,10 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 # and opt-in dependencies?
 option(TROSSEN_ENABLE_REALSENSE "Enable RealSense camera support" ON)
 option(TROSSEN_ENABLE_ZED "Enable StereoLabs ZED camera support" OFF)
+# OFF by default: rerun-cpp adds ~50 MB of download and a multi-minute first build, and is
+# only needed by the stationary-demo ReRun viewer flow. Enable explicitly when building
+# that demo: cmake -DTROSSEN_ENABLE_RERUN_OBSERVER=ON ...
+option(TROSSEN_ENABLE_RERUN_OBSERVER "Build RerunObserver (depends on rerun-cpp SDK)" OFF)
 
 add_library(trossen_sdk SHARED
   src/backends/backend_registry.cpp
@@ -78,6 +82,10 @@ if(TROSSEN_ENABLE_ZED)
   target_link_libraries(trossen_sdk PUBLIC ${ZED_LIBRARIES} ${CUDA_LIBRARIES})
 endif()
 
+# RerunObserver wiring is deferred until after Arrow is found below: rerun-cpp needs
+# Arrow::arrow_shared to satisfy its RERUN_ARROW_LINK_SHARED=ON build mode, and we reuse
+# the same Arrow trossen_sdk already links against to avoid pulling in a second build.
+
 # Add architecture-specific CMAKE_PREFIX_PATH
 if(CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64")
   set(CMAKE_PREFIX_PATH ${CMAKE_PREFIX_PATH} /usr/lib/x86_64-linux-gnu/cmake)
@@ -116,6 +124,70 @@ find_package(Parquet REQUIRED)
 find_package(libtrossen_arm REQUIRED)
 
 include(FetchContent)
+
+# --- ReRun C++ SDK (RerunObserver) ------------------------------------------
+# ReRun is fetched on demand via FetchContent. We reuse the system Arrow that
+# trossen_sdk already links against by passing RERUN_DOWNLOAD_AND_BUILD_ARROW=OFF +
+# RERUN_ARROW_LINK_SHARED=ON, so rerun_sdk does NOT pull in a second, conflicting Arrow.
+# Linking is PRIVATE: the rerun headers stay confined to rerun_observer.cpp.
+if(TROSSEN_ENABLE_RERUN_OBSERVER)
+  message(STATUS "RerunObserver enabled - fetching rerun-cpp SDK")
+
+  set(RERUN_SDK_VERSION "0.32.0")
+  set(RERUN_SDK_URL
+    "https://github.com/rerun-io/rerun/releases/download/${RERUN_SDK_VERSION}/rerun_cpp_sdk.zip")
+  set(RERUN_SDK_HASH
+    "cabd6619b8ec65636ec98767309829c9bdbae1be3069215bd7ad7b15525c8dda")
+
+  set(RERUN_DOWNLOAD_AND_BUILD_ARROW OFF CACHE BOOL
+    "Reuse system Arrow already found by trossen_sdk" FORCE)
+  set(RERUN_ARROW_LINK_SHARED ON CACHE BOOL
+    "Link rerun_sdk against shared Arrow (matches Arrow::arrow_shared)" FORCE)
+
+  # trossen_sdk is SHARED, so every static archive it consumes must be PIC. rerun_sdk's
+  # CMake does not set this on its own target, so we set the project-wide default before
+  # bringing it in.
+  set(_TROSSEN_PREV_PIC ${CMAKE_POSITION_INDEPENDENT_CODE})
+  set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+
+  FetchContent_Declare(
+    rerun_sdk
+    URL ${RERUN_SDK_URL}
+    URL_HASH SHA256=${RERUN_SDK_HASH}
+    DOWNLOAD_EXTRACT_TIMESTAMP TRUE
+  )
+  FetchContent_MakeAvailable(rerun_sdk)
+
+  if(TARGET rerun_sdk)
+    set_target_properties(rerun_sdk PROPERTIES POSITION_INDEPENDENT_CODE ON)
+  endif()
+
+  set(CMAKE_POSITION_INDEPENDENT_CODE ${_TROSSEN_PREV_PIC})
+
+  target_sources(trossen_sdk PRIVATE src/observer/rerun_observer.cpp)
+  target_compile_definitions(trossen_sdk PRIVATE TROSSEN_ENABLE_RERUN_OBSERVER)
+  target_link_libraries(trossen_sdk PRIVATE rerun_sdk)
+
+  # Both librerun_c and libfoxglove bundle the Rust standard library; each archive carries
+  # its own copy of personality / panic handler symbols. --allow-multiple-definition lets
+  # the linker pick one duplicate-symbol copy without error - the documented workaround
+  # for combining multiple Rust static libs in a single C++ shared library.
+  #
+  # TODO(observer-rerun): produce a symbol-collision audit (`nm libtrossen_sdk.so | sort
+  # | uniq -d`) and exercise the Rust panic path under ASan to validate this is safe.
+  # Tracked in the ADR-003 follow-up board. Gated behind TROSSEN_ENABLE_RERUN_OBSERVER
+  # (default OFF), so default builds are unaffected.
+  #
+  # NOTE: an earlier draft also passed `--exclude-libs=ALL` to hide the Rust runtime
+  # symbols. That flag is too broad - it hides every static archive trossen_sdk links
+  # (SCServo, trossen_slate, foxglove, protobuf) from the exported interface and would
+  # silently break downstream consumers that resolve symbols transitively. Reintroduce
+  # only with a narrow whitelist (e.g. --exclude-libs=librerun_c.a).
+  if(UNIX AND NOT APPLE)
+    target_link_options(trossen_sdk PRIVATE
+      "LINKER:--allow-multiple-definition")
+  endif()
+endif()
 
 set(FOXGLOVE_VERSION "0.16.1")
 # Detect platform for Foxglove SDK download
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -172,6 +172,9 @@ if(TROSSEN_ENABLE_RERUN_OBSERVER)
   # its own copy of personality / panic handler symbols. --allow-multiple-definition lets
   # the linker pick one duplicate-symbol copy without error - the documented workaround
   # for combining multiple Rust static libs in a single C++ shared library.
+  # See: https://github.com/rust-lang/rust/issues/44322 (canonical rust-lang issue on
+  #   duplicate runtime symbols from multiple staticlibs)
+  # See: https://sourceware.org/binutils/docs/ld/Options.html (GNU ld --allow-multiple-definition)
   #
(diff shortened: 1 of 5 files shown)


shantanuparab-tr:
* **#245** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/245?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#244** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/244?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#243** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/243?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#242** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/242?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/242?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#241** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/241?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#240** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/240?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#239** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/239?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#238** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/238?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

CMakeLists.txt
```diff
@@ -117,6 +125,70 @@ find_package(libtrossen_arm REQUIRED)
...
+  set(CMAKE_POSITION_INDEPENDENT_CODE ${_TROSSEN_PREV_PIC})
+
+  target_sources(trossen_sdk PRIVATE src/observer/rerun_observer.cpp)
+  target_compile_definitions(trossen_sdk PRIVATE TROSSEN_ENABLE_RERUN_OBSERVER)
+  target_link_libraries(trossen_sdk PRIVATE rerun_sdk)
+
+  # Both librerun_c and libfoxglove bundle the Rust standard library; each archive carries
+  # its own copy of personality / panic handler symbols. --allow-multiple-definition lets
+  # the linker pick one duplicate-symbol copy without error - the documented workaround
+  # for combining multiple Rust static libs in a single C++ shared library.
```
lukeschmitt-tr: Please provide a citation for this

src/observer/rerun_observer.cpp
```diff
@@ -0,0 +1,233 @@
...
+constexpr const char* kDefaultRerunUrl = "rerun+http://127.0.0.1:9876/proxy";
+constexpr const char* kDefaultAppId = "trossen_sdk";
+
+}  // namespace
+
+namespace detail {
+
+std::vector<uint8_t> mat_to_rgb_bytes(const cv::Mat& image, const std::string& encoding) {
+  cv::Mat rgb;
+  if (encoding == "rgb8") {
```
lukeschmitt-tr, line 34: We run this check per-record when the encoding won't change. Can we save it and use a switch/map?

Worse case (mono8), we're doing this comparison three times per record.

src/observer/rerun_observer.cpp
```diff
@@ -0,0 +1,233 @@
...
+      return {};
+    }
+    cv::cvtColor(image, rgb, cv::COLOR_GRAY2RGB);
+  } else {
+    return {};
+  }
+  const std::size_t byte_count =
+    static_cast<std::size_t>(rgb.total()) * static_cast<std::size_t>(rgb.elemSize());
+  std::vector<uint8_t> bytes(byte_count);
+  std::memcpy(bytes.data(), rgb.data, byte_count);
```
lukeschmitt-tr: Is there a way we can avoid a memcpy?

rerun has a non-owning Image constructor - thoughts on using that?

https://ref.rerun.io/docs/cpp/stable/structrerun_1_1archetypes_1_1Image.html#a8b125f8c830483fe36cd5efb150879c2

src/observer/rerun_observer.cpp
```diff
@@ -0,0 +1,233 @@
...
+    }
+    if (!js->efforts.empty()) {
+      std::vector<double> eff(js->efforts.begin(), js->efforts.end());
+      rec_->log(record_id + "/efforts", rerun::Scalars(std::move(eff)));
+    }
+    return;
+  }
+
+  skip("unsupported_record_type");
+}
```
lukeschmitt-tr, line 317: Missing odom once we add the mobile

src/observer/rerun_observer.cpp
```diff
@@ -0,0 +1,233 @@
...
+        skip("unsupported_depth_format");
+        return;
+      }
+      const auto width = static_cast<uint32_t>(depth.cols);
+      const auto height = static_cast<uint32_t>(depth.rows);
+      // rerun-cpp's with_meter expects depth units per meter; depth_scale is the
+      // reciprocal (meters per unit).
+      const float meter = 1.0f / depth_scale;
+      rec_->log(
+        record_id + "/depth",
```
lukeschmitt-tr: We're doing a lot of string concatenation - can we cache/pre-compute these somehow?

lukeschmitt-tr (changes requested):

CMakeLists.txt
shantanuparab-tr: Added citation in the comment block — links to https://github.com/rust-lang/rust/issues/44322 (the canonical rust-lang issue on duplicate runtime symbols when linking multiple staticlibs) and the GNU ld manual entry for `--allow-multiple-definition`.

src/observer/rerun_observer.cpp
shantanuparab-tr: Added an `Odometry2DRecord` dispatch arm — logs pose (x/y/theta) and body-frame twist (linear_x/linear_y/angular_z) as `rerun::Scalars` timelines under `<record_id>/pose/*` and `<record_id>/twist/*`, mirroring the JointState arm. Slate-base subscriptions now render in rerun without the mobile PR having to come back and re-edit this file.

src/observer/rerun_observer.cpp
shantanuparab-tr: Encoding is stamped once per stream by every image producer (confirmed in opencv/realsense/mock_*), so caching the resolved form is safe. Added `detail::ColorEncoding` enum + `detail::resolve_encoding()`. `mat_to_rgb_bytes` now switches on the enum (exhaustive, no default — adding a value triggers `-Wswitch`). A new `PerSubscriptionState` struct holds the encoding (lazy-resolved on first record per `record_id`, sticky on `kUnsupported`). Tests migrated to the enum signature + new `resolve_encoding` coverage test.

src/observer/rerun_observer.cpp
shantanuparab-tr: Cached the suffixed paths on `PerSubscriptionState`. Three nested structs — `ImagePaths{depth}`, `JointPaths{positions,velocities,efforts}`, `OdomPaths{6 fields}` — each in `std::optional<>` and lazy-populated at the top of the matching `dynamic_cast` arm. Replaces six odom concats, three joint concats, and one image-depth concat per frame. Image-color path is bare `record_id`, no caching needed. Joint state caches 3 paths (not per-joint) because the existing code logs bulk `Scalars` vectors, so joint-count variability is irrelevant.

src/observer/rerun_observer.cpp
shantanuparab-tr: Switched to the exact `Image(const T*, WidthHeight, ColorModel)` ctor you linked. `std::vector<uint8_t>` + `std::memcpy` are gone; `mat_to_rgb` returns `cv::Mat` and the call site borrows `rgb.data` directly. One structural Arrow copy remains (unavoidable — traced through rerun-cpp 0.32: the templated ctor wraps the pointer with `Collection<uint8_t>::borrow` and synchronously serialises via `ComponentBatch::from_loggable`, so the borrow ends when the ctor returns). `SAFETY:` comment at the borrow site pins the rerun-cpp 0.32 invariant. Added `non_contiguous_rgb` skip reason (mirrors the depth path's continuity check) and a `ImageBorrow_OutlivesSourceMat` lifetime stress test (caveat noted: ASan-only-meaningful).

include/trossen_sdk/observer/rerun_observer.hpp
```diff
@@ -0,0 +1,127 @@
...
+ *  - ``data::ImageRecord``       -> ``rerun::Image`` (and ``rerun::DepthImage`` when present)
+ *  - ``data::JointStateRecord``  -> three ``rerun::Scalars`` paths (positions / velocities /
+ *                                   efforts)
+ *  - other record types are silently skipped.
+ *
+ * Lifecycle follows ADR-003 §5 L2: ``on_start()`` opens the gRPC connection; ``on_stop()``
+ * tears the stream down. A failed gRPC connect returns ``false`` from ``on_start()`` so the
+ * surrounding ``SessionManager`` can mark the observer dead and continue recording.
+ */
+class RerunObserver : public ObserverBase {
```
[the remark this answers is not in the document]
shantanuparab-tr: Will follow the existing `TROSSEN_ENABLE_REALSENSE`-guarded pattern in the umbrella (a `#ifdef TROSSEN_ENABLE_RERUN_OBSERVER` include block) so the unconditional path stays clean. Tracking as a follow-up — users today reach `RerunObserver` via `ObserverRegistry::create("rerun", ...)`, so the umbrella gap is discoverability, not access.

lukeschmitt-tr (approved):

shantanuparab-tr:
### Merge activity

* **Jun 1, 11:47 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/242).
* **Jun 1, 11:55 PM UTC**: [Graphite](https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/242) rebased this pull request as part of a merge.
* **Jun 1, 11:55 PM UTC**: @shantanuparab-tr merged this pull request with [Graphite](https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/242).

PR #245, opened 2026-05-28
docs: document observers config and rerun live-viewer setup

shantanuparab-tr:
## Summary

- Add an `Observers` section to `docs/configuration.rst` covering the config schema, subscription fields (including the `fields` filter), the ReRun-specific `rerun_url` / `app_id` keys, and the `-DTROSSEN_ENABLE_RERUN_OBSERVER=ON` build gate.
- Add a `Streaming with ReRun` section to `docs/visualize.rst` covering prerequisites, viewer launch, observer config, and across-episodes behavior (entity-tree reset + wall-clock timeline).
- Cross-reference observers from `docs/record.rst` (Recording phase + What's Next).

Built on top of #244 (`05-22-rerun-hw-hardening`); merge after the rerun observer code lands.

## Test plan

- [x] `make docs-strict` builds without warnings (requires `make docs-deps` first if Sphinx isn't installed locally).
- [x] `make docs-serve` and visually verify the new Observers and Streaming with ReRun sections render correctly with intact cross-refs.
- [x] Confirm no mention of the `spawn` config key anywhere (the option exists in code but is currently non-functional).

PR #246, opened 2026-06-02
control: per-episode arm staging to a home pose and hold-pose on teleop stop

shantanuparab-tr:
## Summary

Rebased onto current `main` (the observer work this branch was originally stacked on has since landed via #238–#245). Three changes remain, all surfacing through the three AI example apps: an opt-in per-episode arm-staging flow that moves each arm to a configured home pose at the start of every episode, a teleop-stop fix so the arm holds its pose on Ctrl+C instead of dropping under gravity, and removal of stale scratch demos.

## Changes

- In `include/trossen_sdk/configuration/types/hardware/arm_config.hpp` + `src/...`: carry `staged_position` and `teleop_moving_time_s` through `from_json`/`to_json`. `to_json()` was dropping both, so the home pose declared in `config.json` never reached `TrossenArmComponent` and staging silently no-op'd. `staged_position` is omitted when empty so the no-staging case stays valid.
- In `include/trossen_sdk/configuration/sdk_config.hpp` + `src/configuration/sdk_config.cpp`: add an opt-in, default-off `SdkConfig.stage_each_episode`. When set, each example stages every arm with a configured `staged_position` at the start of *every* episode; when unset, no staging happens. The shipped example configs set it `false`.
- In `include/trossen_sdk/hw/teleop/teleop_controller.hpp` + `src/hw/teleop/teleop_controller.cpp`: add `pause_mirror()`, which stops the control thread without tearing down the drivers, so staging can cooperate with the continuous mirror loop instead of fighting it. Also drops the one-shot `stage()` that ran in the constructor — staging now lives entirely in the application episode loop.
- In `src/hw/arm/trossen_arm_component.cpp`: on teleop stop, switch straight into position mode and command the current measured pose first (entering position mode locks the arm where it is, matching the `trossen_arm` teleop demo's safe-stop sequence), then move to rest over the configured trajectory time. Previously `end_teleop()` switched to idle mode first, dropping motor torque and letting the arm fall under gravity before position control re-engaged.
- In `examples/trossen_{solo,stationary,mobile}_ai/{trossen_*_ai.cpp,config.json,README.md}`: episode loop now pauses the mirror, stages the arms (blocking move to home), then `prepare_teleop()` re-arms teleop modes before `teleop()` restarts the loop. Shipped configs set `stage_each_episode` false; READMEs document the opt-in flag.
- In `examples/` + `examples/CMakeLists.txt`: remove the `callback_demo`, `realsense_rgb` config-only sample, and the backend/hardware/producer registry demos along with their CMake targets. These were scratch demos; the three AI example apps remain the supported entry points.

## Test Plan

- [ ] Builds cleanly (`make build`)
- [ ] Tests pass (`make test`) — `test_config`, `test_teleop_controller`
- [ ] Lint passes (`pre-commit run --all-files`)
- [ ] Tested on hardware — with `stage_each_episode` enabled, each arm moves to its configured home pose at the start of every episode; on Ctrl+C the arm holds its pose (no gravity drop) before returning to rest

## Breaking Changes

No API or config breaks: the new config keys (`stage_each_episode`, `staged_position`, `teleop_moving_time_s`) are optional and default off, so existing configs load unchanged. The teleop-stop sequence is a deliberate runtime behavior change — the arm now holds position on stop instead of going briefly limp.

## Related Issues

Relates to TDS-188


src/hw/teleop/teleop_controller.cpp
```diff
@@ -124,6 +119,16 @@ void TeleopController::reset_teleop() {
   }
 }
 
+void TeleopController::pause_mirror() {
+  // Stop the mirror thread but keep the drivers alive. running_ is left false
+  // so the next prepare_teleop() re-arms teleop modes and teleop() can restart
+  // the loop. Unlike stop_teleop(), end_teleop() is not called.
+  running_.store(false);
```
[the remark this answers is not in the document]
shantanuparab-tr: Good catch — confirmed and fixed. The mirror loop is meant to run continuously across episodes, so the happy path early-returns on `running_.exchange(true)` and never reassigns `thread_`. But `control_loop()`'s internal catch sets `running_=false` *without* joining, and the only per-episode reaper (`pause_mirror()`) is called solely when `stage_each_episode` is true — so under the default config a loop that exits via exception leaves `thread_` joinable, and the next episode's `teleop()` reassigns it → `std::terminate`.

Fixed in `teleop()` by joining any stale, non-running thread before starting a new one (will be in the next push):

```cpp
void TeleopController::teleop() {
  if (running_.exchange(true)) {
    return;
  }
  // Reap a previous loop that exited on its own: control_loop() catches and
  // sets running_=false on exception but does not join, so the finished thread
  // stays joinable. Assigning to a joinable std::thread calls std::terminate,
  // so join any stale thread before starting a new one.
  if (thread_.joinable()) {
    thread_.join();
  }
  thread_ = std::thread([this]() { control_loop(); });
}
```

include/trossen_sdk/configuration/types/hardware/arm_config.hpp
```diff
@@ -32,20 +35,42 @@ struct ArmConfig {
   /// @brief End effector type (e.g. "wxai_v0_follower", "wxai_v0_leader")
   std::string end_effector{"wxai_v0_follower"};
 
+  /// @brief Joint-space home pose the arm moves to when staged. Empty disables
+  /// staging. Length must match the arm's joint count (validated downstream by
+  /// TrossenArmComponent::configure()).
+  std::vector<float> staged_position;
+
+  /// @brief Trajectory time (seconds) used by staging and the rest move.
+  float teleop_moving_time_s{2.0f};
```
lukeschmitt-tr, line 44: If this is only used for staging, we should name this variable more appropriately - right now sounds like it's only related to teleop

include/trossen_sdk/hw/teleop/teleop_controller.hpp
```diff
@@ -89,6 +89,20 @@ class TeleopController {
...
+   * Joins the control thread but, unlike stop_teleop(), does NOT call
+   * end_teleop() — the drivers stay configured and the arms hold their last
+   * commanded pose. After this, the arms can be re-staged, and a subsequent
+   * prepare_teleop() re-arms teleop modes (running_ is false again) followed
+   * by teleop() to restart the loop. Idempotent; no-op if not running.
+   *
+   * Used to re-stage arms to their home pose between episodes without tearing
+   * down teleop.
+   */
+  void pause_mirror();
```
lukeschmitt-tr, line 104: Can we call this similarly to other methods? How's `pause_teleop`?

include/trossen_sdk/hw/teleop/teleop_controller.hpp
```diff
@@ -89,6 +89,20 @@ class TeleopController {
    */
   void reset_teleop();
 
+  /**
+   * @brief Pause the mirror loop without releasing the hardware.
+   *
+   * Joins the control thread but, unlike stop_teleop(), does NOT call
+   * end_teleop() — the drivers stay configured and the arms hold their last
+   * commanded pose. After this, the arms can be re-staged, and a subsequent
```
lukeschmitt-tr, line 97: This comment is very specific about arms - make it more general to hardware

examples/trossen_mobile_ai/config.json
```diff
@@ -1,6 +1,8 @@
 {
   "robot_name": "trossen_mobile_ai",
 
+  "stage_each_episode": false,
```
lukeschmitt-tr, line 4: Does it make sense for this to be a top level config option?

lukeschmitt-tr (changes requested):

shantanuparab-tr:
Closed in favor of #248

PR #248, opened 2026-06-09
control: per-episode hardware lifecycle driven by SessionManager

shantanuparab-tr:
## Summary

Moves the per-episode teleop + staging choreography out of the example apps and into the SDK: the `SessionManager` now owns a per-episode hardware lifecycle, components opt in and implement their own per-episode work, and the three AI examples collapse to a single `add_teleop()` registration. Supersedes #246.

## Changes

- In `include/trossen_sdk/hw/hardware_component.hpp`: add opt-in, no-op per-episode lifecycle hooks `on_pre_episode()` / `on_episode_started()` / `on_episode_ended()` plus a virtual `is_episode_lifecycle_enabled()` (default `false`). The base parses no config — each component answers the getter from its own config. The per-episode lifecycle runs on both a clean stop and a discard/re-record (user episode-ended *callbacks* stay clean-stop only); a component only participates if it backs a registered producer.
- In `include/trossen_sdk/hw/producer_base.hpp` + `src/runtime/producer_registry.cpp` + `src/runtime/push_producer_registry.cpp`: producers retain their backing `HardwareComponent` and expose it via `hardware()`; the registries set it once after construction, so the `SessionManager` can reach every component's hooks with no per-producer code.
- In `include/trossen_sdk/runtime/session_manager.hpp` + `src/runtime/session_manager.cpp`: add `add_teleop()` + a `teleop_controllers_` list and `collect_lifecycle_components_()` (the unique, opted-in, producer-backed components). Drive the per-episode bracket — pause mirrors → `on_pre_episode()` (arms re-home) → re-arm in the pre-episode phase; `teleop()` + `on_episode_started()` once recording is live; `reset_teleop()` + `on_episode_ended()` on episode end; `stop_teleop()` at shutdown. Opted-in components stage **in parallel** (independent hardware via `std::async`), so the episode-boundary wait is bounded by the slowest arm rather than the sum. A staging hook that throws is caught and routed through the same `cleanup_and_abort()` path as every other `start_episode()` failure.
- In `include/trossen_sdk/hw/arm/trossen_arm_component.hpp` + `src/hw/arm/trossen_arm_component.cpp`: override `on_pre_episode()` → `stage()`, parse `episode_lifecycle_enabled`, and override `is_episode_lifecycle_enabled()`. Rename `teleop_moving_time_s` → `staging_time_s` (it governs any SDK-commanded point-to-point move — staging to home and the return-to-rest move — not just teleop). On teleop stop, hold the measured pose before returning to rest so the arm doesn't drop under gravity.
- In `include/trossen_sdk/hw/teleop/teleop_controller.hpp` + `src/hw/teleop/teleop_controller.cpp`: rename `pause_mirror()` → `pause_teleop()` for naming consistency, and join a stale control thread on restart so repeated `teleop()` across episodes cannot `std::terminate`.
- In `include/trossen_sdk/configuration/types/hardware/arm_config.hpp`: carry `staged_position`, `staging_time_s`, and `episode_lifecycle_enabled` through `from_json`/`to_json`.
- In `examples/trossen_{solo,stationary,mobile}_ai/{trossen_*_ai.cpp,config.json,README.md}`: hand controllers to the manager via `add_teleop()` and remove the per-episode pause/stage/re-arm, mirror-start, reset, and stop choreography the manager now owns. Example configs stay opt-out (no `episode_lifecycle_enabled`); each README's "Optional: home staging" section documents how to enable staging per arm.

## Test Plan

- [x] Builds cleanly (`make build`)
- [x] Tests pass — `test_config`, `test_teleop_controller`, and the session-manager suite
- [x] Lint passes (`pre-commit` on changed files — cpplint, codespell, json, whitespace)
- [x] Tested on hardware — with staging enabled, opted-in arms re-home at the start of every episode; on Ctrl+C the arm holds its pose (no gravity drop) before returning to rest

## Breaking Changes

The new lifecycle fields — `episode_lifecycle_enabled`, `staged_position`, and `staging_time_s` — are optional and default off/empty, so configs that don't set them load unchanged. One existing key is **renamed, not aliased**: `teleop_moving_time_s` → `staging_time_s`. A config still using the old `teleop_moving_time_s` key will have it silently ignored and fall back to the `2.0s` default; update such configs to `staging_time_s`. This is a deliberate clean break (pre-1.0, no config-compat promise), not a dual-key alias.

ABI is not preserved — `HardwareComponent` gains virtuals and the producer bases gain a member; downstream consumers rebuild on upgrade (pre-1.0, no ABI promise). The teleop-stop sequence is a deliberate runtime behavior change — the arm holds position on stop instead of going briefly limp.

## Related Issues

Relates to TDS-188

### DIFF
--- a/examples/trossen_mobile_ai/config.json
+++ b/examples/trossen_mobile_ai/config.json
@@ -1,6 +1,7 @@
 {
   "robot_name": "trossen_mobile_ai",
 
+
   "hardware": {
     "arms": {
       "leader_left": {
--- a/examples/trossen_mobile_ai/config.json
+++ b/examples/trossen_mobile_ai/config.json
@@ -1,7 +1,6 @@
 {
   "robot_name": "trossen_mobile_ai",
 
-
   "hardware": {
     "arms": {
       "leader_left": {
--- a/include/trossen_sdk/configuration/types/hardware/arm_config.hpp
+++ b/include/trossen_sdk/configuration/types/hardware/arm_config.hpp
@@ -22,7 +22,7 @@ namespace trossen::configuration {
  *   "model": "wxai_v0",
  *   "end_effector": "wxai_v0_leader",
  *   "staged_position": [0.0, 1.04, 0.52, 0.63, 0.0, 0.0, 0.0],  // optional
- *   "slew_time_s": 2.0,                                   // optional
+ *   "staging_time_s": 2.0,                                // optional
  *   "episode_lifecycle_enabled": true                     // optional, default false
  * }
  */
@@ -41,12 +41,12 @@ struct ArmConfig {
   /// TrossenArmComponent::configure()).
   std::vector<float> staged_position;
 
-  /// @brief Slew time (seconds): the duration over which an SDK-commanded
+  /// @brief Staging time (seconds): the duration over which an SDK-commanded
   /// point-to-point move runs (staging to home and the return-to-rest move).
   /// Longer = slower, gentler motion. Sized so that a large start-to-goal
   /// difference does not exceed joint velocity limits or produce violent
   /// motion. Optional; 2.0s is a safe default for the supported arms.
-  float slew_time_s{2.0f};
+  float staging_time_s{2.0f};
 
   /// @brief Whether this arm participates in the SessionManager's per-episode
   /// lifecycle (staging to its home pose before each episode). Opt-in; defaults to
@@ -61,8 +61,8 @@ struct ArmConfig {
     if (j.contains("staged_position")) {
       j.at("staged_position").get_to(c.staged_position);
     }
-    if (j.contains("slew_time_s")) {
-      j.at("slew_time_s").get_to(c.slew_time_s);
+    if (j.contains("staging_time_s")) {
+      j.at("staging_time_s").get_to(c.staging_time_s);
     }
     if (j.contains("episode_lifecycle_enabled")) {
       j.at("episode_lifecycle_enabled").get_to(c.episode_lifecycle_enabled);
@@ -75,7 +75,7 @@ struct ArmConfig {
       {"ip_address", ip_address},
       {"model", model},
       {"end_effector", end_effector},
-      {"slew_time_s", slew_time_s},
+      {"staging_time_s", staging_time_s},
       {"episode_lifecycle_enabled", episode_lifecycle_enabled}
     };
     // Emit staging only when configured. TrossenArmComponent::configure()
--- a/include/trossen_sdk/configuration/types/hardware/arm_config.hpp
+++ b/include/trossen_sdk/configuration/types/hardware/arm_config.hpp
@@ -7,6 +7,7 @@
 #define TROSSEN_SDK__CONFIGURATION__TYPES__HARDWARE__ARM_CONFIG_HPP_
 
 #include <string>
+#include <vector>
 
 #include "nlohmann/json.hpp"
 
@@ -19,7 +20,10 @@ namespace trossen::configuration {
  * {
  *   "ip_address": "192.168.1.3",
  *   "model": "wxai_v0",
- *   "end_effector": "wxai_v0_leader"
+ *   "end_effector": "wxai_v0_leader",
+ *   "staged_position": [0.0, 1.04, 0.52, 0.63, 0.0, 0.0, 0.0],  // optional
+ *   "slew_time_s": 2.0,                                   // optional
+ *   "episode_lifecycle_enabled": true                     // optional, default false
  * }
  */
 struct ArmConfig {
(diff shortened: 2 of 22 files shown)


shantanuparab-tr:
* **#283** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/283?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#282** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/282?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#281** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/281?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#255** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/255?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#254** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/254?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#253** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/253?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#252** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/252?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#251** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/251?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#250** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/250?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#249** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/249?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#248** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/248?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/248?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

examples/trossen_mobile_ai/config.json
```diff
@@ -1,6 +1,7 @@
 {
   "robot_name": "trossen_mobile_ai",
 
+
```
lukeschmitt-tr:
```suggestion
```

include/trossen_sdk/configuration/types/hardware/arm_config.hpp
```diff
@@ -32,20 +36,55 @@ struct ArmConfig {
...
+  /// staging. Length must match the arm's joint count (validated downstream by
+  /// TrossenArmComponent::configure()).
+  std::vector<float> staged_position;
+
+  /// @brief Slew time (seconds): the duration over which an SDK-commanded
+  /// point-to-point move runs (staging to home and the return-to-rest move).
+  /// Longer = slower, gentler motion. Sized so that a large start-to-goal
+  /// difference does not exceed joint velocity limits or produce violent
+  /// motion. Optional; 2.0s is a safe default for the supported arms.
+  float slew_time_s{2.0f};
```
lukeschmitt-tr: I don't think slew time makes much sense as the config name here. Don't we already have something like this? teleop time or similar? staging_time_s might also work

include/trossen_sdk/hw/teleop/teleop_controller.hpp
```diff
@@ -89,6 +89,20 @@ class TeleopController {
...
+   * Joins the control thread but, unlike stop_teleop(), does NOT call
+   * end_teleop() — the drivers stay configured and the robots hold their last
+   * commanded pose. After this, the robots can be re-staged, and a subsequent
+   * prepare_teleop() re-arms teleop modes (running_ is false again) followed
+   * by teleop() to restart the loop. Idempotent; no-op if not running.
+   *
+   * Used to re-stage robots to their home pose between episodes without tearing
+   * down teleop.
+   */
+  void pause_teleop();
```
lukeschmitt-tr, line 104: How is this different from reset_teleop?

include/trossen_sdk/hw/hardware_component.hpp
```diff
@@ -62,6 +62,67 @@ class HardwareComponent {
...
+   * the hooks.
+   *
+   * @note The SessionManager discovers lifecycle components through the producers
+   *       registered with it, so a component participates only if it also backs at
+   *       least one producer added via add_producer()/add_push_producer(). Enabling the
+   *       flag on a component that is never recorded has no effect.
+   *
+   * @return true if this component participates in the per-episode lifecycle
+   */
+  virtual bool is_episode_lifecycle_enabled() const { return false; }
```
lukeschmitt-tr, line 124: Looks like we're adding this getter but not the underlying `episode_lifecycle_enabled_` flag to this class? Is this assessment correct?

lukeschmitt-tr (changes requested):

include/trossen_sdk/configuration/types/hardware/arm_config.hpp
shantanuparab-tr: We had teleop time but it wasn't fitting well as it wasn't actually used for teleoperation. I'll change it to staging

include/trossen_sdk/hw/teleop/teleop_controller.hpp
shantanuparab-tr: reset_teleop() calls  post_episode() and keeps the mirror loop running (reposition the leader while recording is stopped);  pause_teleop() joins the control thread (stops the mirror) but does not call end_teleop(), so drivers  stay armed for re-staging between episodes.

include/trossen_sdk/hw/hardware_component.hpp
shantanuparab-tr: The base has no member;  TrossenArmComponent holds episode_lifecycle_enabled_ and parses it from its own config. This is the deliberate "base parses no config; each   component answers from its own config" pattern, already documented in the virtual's doc comment.

lukeschmitt-tr (approved):

PR #282, opened 2026-07-28
build(docker): fix the image build and update the driver pin

shantanuparab-tr:
Stacked on #281.

- `libtrossen_arm` 1.10.0, as a build argument (`--build-arg TROSSEN_ARM_VERSION=`). The
  driver major.minor must match the arm controller firmware; v1.11.0 is available.
- Installs `libfastcdr-dev`, `libfastrtps-dev` and `librealsense2-dev`, required by the
  `TROSSEN_ENABLE_REALSENSE=ON` configure step that `RUN make` performs.
- `WORKDIR /app/build/examples`, the directory holding the example binaries.
- `.dockerignore` excludes `build*/` (was `build/`) and `.claude/`.

Verification: `docker build -f Dockerfile .` succeeds and the default command runs.

```
$ docker run --rm trossen-sdk:after
Registered producer type: trossen_arm
Registered push producer type: realsense_camera
Usage: ./trossen_solo_ai [OPTIONS]
```

PR #281, opened 2026-07-28
ci: cover both policy-client gate states; fix cwd-relative test fixture path

shantanuparab-tr:
Stacked on #255. Builds and tests the SDK across every CMake flag configuration.

## CI

`.github/ci-configs.json` holds the build configurations; `select-configs` emits them into
`strategy.matrix`:

| Config | Flags | Tests |
| --- | --- | --- |
| `default` | `REALSENSE=ON` | 391 |
| `policy` | `+ POLICY_CLIENT=ON` | 478 |
| `policy-no-realsense` | `REALSENSE=OFF POLICY_CLIENT=ON` | 461 |
| `minimal` | `REALSENSE=OFF` | 374 |
| `policy-no-testing` | `+ BUILD_TESTING=OFF` | none |
| `policy-release` | `+ CMAKE_BUILD_TYPE=Release` | 478 |
| `policy-python` | `+ BUILD_PYTHON_BINDINGS=ON` | 478 |
| `policy-rerun` | `+ RERUN_OBSERVER=ON` | 500 |
| `all-on` | policy + rerun + pybind | 500 |

- Pull requests build the configurations marked `pr` (`default`, `policy`). Pushes to `main`, `schedule` and `workflow_dispatch` build all nine.
- Pull requests against any base branch trigger the workflow.
- `schedule` is `cron: '37 10 * * 1'`, Mondays 10:37 UTC. GitHub runs scheduled workflows from the default branch only.
- `concurrency` cancels in-progress runs for the same ref.
- New packages: `libgrpc++-dev` (pkg-config `grpc++`), `protobuf-compiler-grpc `(`grpc_cpp_plugin`) and `python3-dev`.
- One `cmake -B build` per job, then build, then `ctest` unless the configuration sets `skip_tests`.
- No configuration sets `TROSSEN_ENABLE_ZED`.

## Tests

- `test_backend_registry` loads `tests/test_config.json` via the build-injected `TROSSEN_TEST_CONFIG_PATH` and asserts the fixture exists.
- `ObserverBase.FirstRecord_IsDispatched` and `ObserverBase.SparseStream_DispatchesAfterIdleGap` wait for the handler through a polling helper with a 5 s timeout and assert dispatch, not latency.

## Verification

All nine configurations built and tested locally from `ci-configs.json` with the test counts above. Out-of-tree build: 8/8 ` BackendRegistryTest`. Twelve consecutive full-suite runs at `ctest -j24`: 0 failures. Host-only configuration `+ZED=ON`: 478 tests.

PR #255, opened 2026-06-15
policy_client: stationary examples (openpi + lerobot)

shantanuparab-tr:
## Summary

Two runnable bimanual examples that exercise the whole stack end-to-end against a live policy server — one per transport family — sharing the same 2-follower + 4-camera setup.

> **Stack:** top of the PolicyClient stack, based on #248. Review bottom-up (#254 → here).

## Changes

- In `examples/trossen_stationary_ai_policy/{trossen_stationary_ai_policy.cpp,config.json,README.md,POLICY_INTEGRATION_NOTES.md}`: openpi example (transport `openpi_ws`, WebSocket) — config, runnable app, and integration notes (server setup, tuning knobs, drain-threshold guidance).
- In `examples/trossen_stationary_ai_lerobot/{trossen_stationary_ai_lerobot.cpp,config.json,README.md}`: the LeRobot sibling (transport `lerobot_grpc`, `async_inference`) — same bimanual setup driven by a LeRobot server, with a `pi05`-ready config.
- In `examples/CMakeLists.txt`: build both example binaries behind `TROSSEN_SDK_ENABLE_POLICY_CLIENT`.
- Example configs use `slew_time_s` (matching #248's arm-config rename).

## Test Plan

- [x] Builds cleanly with `-DTROSSEN_SDK_ENABLE_POLICY_CLIENT=ON`
- [x] Lint passes (`pre-commit run --all-files`)
- [x] Hardware — validated against the live pi05 server: bimanual setup driven by the remote policy, chunk playback smooth across boundaries (`working pi05 config for the live integration`)

## Breaking Changes

None. New examples, gated behind `TROSSEN_SDK_ENABLE_POLICY_CLIENT` (default OFF).

## Related

Based on #248 (example configs depend on its `slew_time_s` arm key). Top of the stack over #249–#254.

PR #254, opened 2026-06-15
policy_client: producer + hardware-registry integration

shantanuparab-tr:
## Summary

Wires `PolicyClient` into the SDK runtime: a `PolicyClientProducer` (so the chunk-aligned command stream is a recordable source), registry placement so it constructs from config, and the small arm change the playback path needs.

> **Stack:** PolicyClient stack, based on #248. Review bottom-up (#253 → here). This PR touches `ArmConfig`/`TrossenArmComponent`, which #248 also changes — see Related.

## Changes

- In `include/trossen_sdk/hw/policy/policy_client_producer.{hpp}` + `src/hw/policy/policy_client_producer.cpp`: `PolicyClientProducer` emitting the commanded rows (with `joint_names` in the stream metadata/features); `REGISTER_PRODUCER` / `REGISTER_HARDWARE` under `"policy_client"`.
- In `include/trossen_sdk/hw/active_hardware_registry.hpp` + `src/hw/active_hardware_registry.cpp`: registry placement so a `policy_client` hardware entry is constructed and configured from JSON.
- In `include/trossen_sdk/configuration/types/producers/producer_config.hpp`: carry the policy-client producer config.
- In `include/trossen_sdk/configuration/types/hardware/arm_config.hpp` + `include/trossen_sdk/hw/arm/trossen_arm_component.{hpp,cpp}`: add `write_moving_time_s` — the per-tick trajectory time passed to `set_all_positions()` in `write_joint()` (0 = apply immediately; non-zero smooths per-tick motion, on top of the Face EMA). Reconciled with #248's arm changes: built on #248's `slew_time_s`/`staged_position`/`episode_lifecycle_enabled`, adding only `write_moving_time_s`.
- In `tests/test_policy_client_producer.cpp`, `tests/test_active_hardware_registry.cpp`, `tests/CMakeLists.txt`: producer metadata + registry construction coverage.
- In `cmake/policy_client.cmake`: add `policy_client_producer.cpp` to the gated sources.

## Test Plan

- [x] Builds cleanly with `-DTROSSEN_SDK_ENABLE_POLICY_CLIENT=ON`
- [x] Tests pass — `test_policy_client_producer`, `test_active_hardware_registry`, `test_config`
- [x] Lint passes (`pre-commit run --all-files`)
- [x] Hardware — exercised end-to-end via the live pi05 integration (see #255)

## Breaking Changes

[...]

PR #283, opened 2026-07-29
build: honor the CMake 4.x policy floor and document policy-client gRPC deps

shantanuparab-tr:
## Why

An external user tried to build the SDK with `-DTROSSEN_SDK_ENABLE_POLICY_CLIENT=ON` and hit three separate failures before getting a successful build. All three were real. This PR fixes the two that belong to us and documents the third.

Their fourth issue — a conda `libcurl.so.4` on `LD_LIBRARY_PATH` shadowing the system one and hiding the `CURL_OPENSSL_4` symbols `libarrow` needs — is deliberately **not** addressed here. It is a property of their shell, and reaching into a user's library path from CMake would be fragile and surprising.

## 1. `CMAKE_POLICY_VERSION_MINIMUM` was set in a way that does nothing

CMake 4.x removed compatibility with `cmake_minimum_required` below 3.5. Two fetched dependencies fall under that floor:

- `SCServo` — `src/CMakeLists.txt:1` declares `2.8.3`
- `msgpack-cxx` — `CMakeLists.txt:1` declares `3.1`

`FetchContent_Declare(SCServo)` already carried `CMAKE_ARGS -D CMAKE_POLICY_VERSION_MINIMUM=3.5`, which reads like the problem was handled. It was not. `CMAKE_ARGS` is an ExternalProject *configure-step* option, and FetchContent only runs the download/update/patch steps before handing the dependency to `add_subdirectory`. The flag was recorded into the download-only sub-build and never applied to anything.

Replaced with a directory-scoped variable set once near the top, which propagates into both dependencies via normal subdirectory inheritance:

```cmake
set(CMAKE_POLICY_VERSION_MINIMUM 3.5)
```

`IXWebSocket` needed no change — it declares the range form `3.4.1...3.17.2`, which is the sanctioned escape from the removal and is accepted as-is.

This was invisible to us because CI and our dev boxes run CMake 3.28, where the pre-4.x behavior applies and nothing errors. CMake 4.x is what current pip and newer distros ship, so this will keep recurring for external users.

## 2. `BUILD_WITH_ROS OFF` did not actually keep ROS out of the build

We configure `trossen_slate` with `BUILD_WITH_ROS OFF`, but that flag does not gate slate's ament lookup. Slate does `find_package(ament_cmake QUIET)` unconditionally (guarded only on `SKBUILD`), and if it finds one, runs `ament_package()` — which parses `package.xml` through Python `catkin_pkg`. A developer with a ROS 2 environment sourced therefore needs `catkin_pkg` installed in whichever interpreter CMake picks, to build a dependency we explicitly asked to skip ROS for.

Added before the slate fetch:

```cmake
set(CMAKE_DISABLE_FIND_PACKAGE_ament_cmake ON)
```

This is the standard CMake mechanism for forcing a `QUIET` lookup to fail. Safe here: we consume slate as a plain CMake target (include dirs + link) and never touch its ament export.

## 3. The policy-client gRPC packages were undocumented

`cmake/policy_client.cmake:50-52` hard-requires `grpc++` via `pkg_check_modules`, plus `protoc` and `grpc_cpp_plugin` via `find_program(... REQUIRED)`. Configure fails outright without them, and neither `README.md` nor `docs/installation.rst` mentioned gRPC anywhere.

Added a **PolicyClient (optional)** section to both, following the existing ZED/RealSense optional-dependency convention:

```bash
sudo apt-get install -y libgrpc++-dev protobuf-compiler-grpc
```

Two details worth stating, both included as notes in the docs:

- The gRPC **runtime** package (`libgrpc++1.x`) is not sufficient — `grpc++.pc` ships only in `libgrpc++-dev`. This is the same runtime-vs-dev trap already documented for RealSense, and it is what the reporting user hit.
- `libgrpc-dev` does **not** need to be installed explicitly; `libgrpc++-dev` depends on it. The user installed it by hand, which was harmless but unnecessary.

Note that CI is already correct here — `.github/workflows/build.yml` installs both packages and `.github/ci-configs.json` builds `POLICY_CLIENT=ON` in 7 of 9 configs. This gap was purely user-facing docs.

## Verification

CMake 4.x behavior could not be checked on a normal dev box (3.28.3), so it was verified against a real CMake 4.1.3 installed in a throwaway venv:

| Check | Result |
| --- | --- |
| Bare `cmake_minimum_required(VERSION 2.8.3)` under CMake 4.1.3 | Fails with `Compatibility with CMake < 3.5 has been removed` — reproduces the report |
| Same, with `set(CMAKE_POLICY_VERSION_MINIMUM 3.5)` in the parent scope | Configures; downgraded to a deprecation warning |
| Range form `3.4.1...3.17.2`, no override | Configures — confirms IXWebSocket needs nothing |
| `find_package(ament_cmake QUIET)` with a stub ament package on `CMAKE_PREFIX_PATH` | Found — reproduces the ROS-sourced trigger |
| Same, with `CMAKE_DISABLE_FIND_PACKAGE_ament_cmake ON` | Not found — fix confirmed |
| Full project configure, `REALSENSE=ON POLICY_CLIENT=ON`, **CMake 4.1.3** | `Configuring done` / `Generating done` |
| Full project configure + `trossen_sdk` build, CMake 3.28.3 | Configures and links; no new warnings (the variable is inert pre-4.x) |

The pre-existing `CMP0048` warnings from SCServo are unchanged — on 3.28 the new variable has no effect, so policies there are still resolved as before.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,10 @@ project(trossen_sdk
 set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
+# SCServo (2.8.3) and msgpack-cxx (3.1) declare a cmake_minimum_required below the 3.5
+# floor that CMake 4.x enforces. Raising the floor here covers both fetched dependencies.
+set(CMAKE_POLICY_VERSION_MINIMUM 3.5)
+
 # TODO: Discuss build options - do we want RealSense ON by default? What about other
 # optional hardware (e.g., Zed cameras)? Where do we draw the line between mandatory
 # and opt-in dependencies?
@@ -268,8 +272,6 @@ FetchContent_Declare(
   SYSTEM
   GIT_REPOSITORY https://github.com/ftservo/FTServo_Linux.git
   GIT_TAG 064a6dbf814477b970959b5b01b2fc9bc594afa3
-  CMAKE_ARGS
-    -D CMAKE_POLICY_VERSION_MINIMUM=3.5
   PATCH_COMMAND ${CMAKE_COMMAND} -E chdir <SOURCE_DIR> bash -c "git apply --check ${CMAKE_SOURCE_DIR}/cmake/patches/SCServo.patch && git apply ${CMAKE_SOURCE_DIR}/cmake/patches/SCServo.patch || echo 'Patch already applied'"
 
 )
@@ -296,6 +298,10 @@ set(BUILD_WITH_ROS OFF CACHE BOOL "Build trossen_slate without ROS" FORCE)
 set(BUILD_DEMOS OFF CACHE BOOL "Skip trossen_slate demo executables" FORCE)
 set(TROSSEN_SLATE_BUILD_PYTHON_BINDINGS OFF CACHE BOOL
   "Skip trossen_slate's pybind module; trossen_sdk provides its own" FORCE)
+# trossen_slate probes for ament_cmake regardless of BUILD_WITH_ROS, and its ament_package()
+# path parses package.xml through Python catkin_pkg. Keep it off so a sourced ROS 2
+# environment does not pull ament into this build.
+set(CMAKE_DISABLE_FIND_PACKAGE_ament_cmake ON)
 
 FetchContent_MakeAvailable(trossen_slate)
 
--- a/README.md
+++ b/README.md
@@ -156,6 +156,18 @@ Required for RealSense cameras. Install the RealSense SDK 2.0 by following the [
 
 > **Note:** The two important packages from the RealSense repository are `librealsense2-dev` and `librealsense2-utils`. `realsense-viewer` only needs `librealsense2-utils`, but `find_package(realsense2)` in CMake needs the headers and CMake config from `librealsense2-dev`. If `realsense-viewer` works but the SDK build fails to find RealSense, it means the runtime is installed but the dev package is missing.
 
+### PolicyClient (optional)
+
+Required only when building with `-DTROSSEN_SDK_ENABLE_POLICY_CLIENT=ON`. The gRPC transport is generated from a vendored proto at build time, so the build needs the gRPC development headers and the C++ `protoc` plugin:
+
+```bash
+sudo apt-get install -y \
+    libgrpc++-dev \
+    protobuf-compiler-grpc
+```
+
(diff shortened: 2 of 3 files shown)


shantanuparab-tr:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/283?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#283** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/283?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/283?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#282** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/282?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#281** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/281?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#255** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/255?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#254** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/254?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#253** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/253?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#252** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/252?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#251** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/251?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#250** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/250?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#249** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/249?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#248** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/248?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #285, opened 2026-08-24
teleop: add a gripper force-feedback channel

shantanuparab-tr:
Adds a reverse channel alongside the forward position mirror: the follower reports its measured gripper effort and the leader renders a reflected force, so the operator feels the grasp. Stacked on #284.

- **Three virtuals on** **`TeleopTypeIO`**, all no-op by default — `renders_gripper_feedback()` (leader, gates the channel), `read_gripper_effort() `(follower, `nullopt` if it cannot measure), `apply_gripper_feedback(float)` (leader).
Hardware that can neither sense nor render gripper force opts out by doing nothing.
- **The controller runs it right after the forward write**, so the reflected force lands on the same tick as the command that caused the grasp rather than one tick behind.

**No behaviour change on its own** — nothing overrides the gate yet, so the condition is always false and the mirror loop does exactly what it did before. The arm-side implementation (the cubic curve and its tuning keys) lands with the passive-leader PR.

Two details in the loop body that are deliberate:

- The leader's `bool` is checked **before** the follower's `read_gripper_effort()`. On a networked arm that read is a driver call, and this runs at the control rate, so short-circuiting keeps the default path free.
- A follower that cannot measure is **skipped**, not treated as zero. Zero is a meaningful value here — it means "not grasping" — so substituting it would relax the leader gripper instead of holding position.

PR #287, opened 2026-08-24
feat(arm): clamp, tolerance and smoothing options on the command path

shantanuparab-tr:
Three opt-in options on `TrossenArmComponent`'s command path, all off by default, plus the filter they need. Hardware-verified on rivet-02 (2026-08-07).

| Config | Where it acts |
| --- | --- |
| `command_position_min` / `_max` | host-side per-joint clamp, `null` = joint unclamped |
| `position` / `velocity` / `effort_tolerance` | pushed to the controller's limit checks |
| `smoothing_*` (5 keys) | one-Euro adaptive low-pass, gripper excluded by default |

`write_joint()` is now clamp → smooth → `set_all_positions`.

**Clamp before filter, to prevent windup.** Either order respects the band, but filtering first lets the filter's state follow the command past the bound, so the output stays pinned after the leader returns inside. Band ±1, leader parked at 3× the bound: 10 ms of stick at the shipped tuning, 175 ms at `beta 0`.

**The clamp bounds are the mechanical limit minus exactly 2°, and the plan had this wrong** — it said 0.035 rad, which is 2.0054° and yields 1.5358/3.1066/2.3212 against the rig's shipped 1.5359/3.1067/2.3213. These bounds hold 2.000000°: π/2 − 2° = 1.5358897, π − 2° = 3.1066861, 3π/4 − 2° = 2.3212879. That is a deliberate 1e-4 rad departure from `rivet_01.json`, so **PR 16 must write these values, not copy the preset**.

**J0 is hand-measured and mirrored** — left −1.1…0.8, right −0.8…1.1, the shoulder-yaw stop keeping the arms off each other and the chassis. Not derivable, must not be unified. Documented in `arm_config.hpp` because D10 dropped the `.rst`.

`one_euro_filter.hpp` came with `TODO: verify citation (this is claude generated)` and credited "Casiez, **Godin** & Vogel". Authors are Casiez, **Roussel** & Vogel (CHI 2012); the implementation itself checks out against theirs — including the speed estimate being taken against the previous *filtered* value, which looks wrong and is what the reference does. Replayed against their `groundTruth.csv`: max deviation 4.0e-06, bounded by the fixture's print precision.

[...]

PR #286, opened 2026-08-24
session_control: add a push-based session-control contract and wire it in

shantanuparab-tr:
Lets hardware drive session state transitions. A source detects operator intent on its own thread and pushes it to the manager, instead of the manager polling for it. Stacked on #285.

- **`SessionControlCapable`** — `start()` / `stop()` plus two installed callbacks. Events are semantic intents (`kStart` = "advance"), not raw inputs, so a source needs to know nothing about the session's phase.
- **`interpret_event(event, phase)`** resolves intent plus phase to a `UserAction`: `kStart` means begin an episode when idle, stop-and-advance while recording, skip the wait while resetting.
- Events **wake the loop** through the existing condvars instead of being discovered on the next poll tick.

Two behavioural decisions worth a reviewer's attention:

**Events coalesce in a single slot rather than queueing.** Human-scale input should not stack up behind slow loop iterations — a queue turns two quick presses into a confusing cascade (start, immediate stop-early, start again). Latest intent wins.

**A session-control stop discards the episode in flight.** It lands wherever the operator happened to be, and downstream a truncated episode is indistinguishable from a complete one, so saving it teaches the wrong thing. Since discarding also ends the episode, a caller still doing `if (is_episode_active()) stop_episode();` finds nothing to do and needs no edit. **Ctrl+C is deliberately unchanged** and still saves the partial, because that is what every existing session does today.

Lifetime is handled by ordering rather than by weak pointers: `attach_session_control()` takes a `shared_ptr` and enforces one source at a time, and `~SessionManager()` detaches before tearing anything else down, so a source's reader thread cannot fire a callback into a half-destructed manager.

The legacy arrow-key polling and the `request_rerecord()` / `signal_reset_complete()` API are untouched, so existing sessions behave exactly as before.

Cherry-picked from `experimental/trossen-vr`, where this was already reviewed, rather than rewritten; the partial-discard behaviour is layered on top.

PR #284, opened 2026-08-24
teleop: add a base velocity space and implement it on the SLATE base

shantanuparab-tr:
Adds a third teleop space so a mobile base runs through the same mirror loop as an arm, and implements it on the SLATE base so the interface ships with a consumer rather than as an empty seam. First PR of the Rivet decomposition stack.

- **`Space::Base`** — one enum value, one `kSpaceDescriptors` row, a 5-line `BaseSpaceTeleop`. The `static_assert` against `Space::Count` keeps table and enum in lockstep.
- **Explicit** **`Base`** **/** **`Count`** **cases in** **`TrossenArmComponent::as_space_io`**, both `nullptr`. Listing them rather than falling through means `-Wswitch` flags the _next_ space added.
- **`base_axis`** — `kLinear, kAngular, kLift, kLateral` plus `get()`, because base commands are variable-length: 2 axes for differential drive, 4 for a holonomic base with a lift, both valid followers for the same leader. `get()` yields 0 for an axis a short vector cannot carry instead of reading past the end.
- **`SlateBaseComponent`** **implements it** — `read()`, `write()`,`prepare_for_teleop()`, `end_teleop()`.

**This PR freezes the axis layout, append-only.** `kLateral` is deliberately not beside `kLinear`: index 1 already means angular in shipped SLATE and VR code, so renumbering would silently swap yaw for strafe on working hardware. If the layout should change, it changes here. Signs are measured, not derived — right-handed frame, +x forward, +z up, every axis positive in the left/up/forward sense — so which raw joystick axis and sign feeds each entry is a per-device measurement each component documents itself.

Interface and SLATE implementation are cherry-picked from `experimental/trossen-vr `(already reviewed there) rather than rewritten. That version keeps `prepare_for_teleop()` motor torque — without it the first writes of a session go nowhere — and stays silent on short commands instead of logging per tick in the mirror loop. Layered on top: `base_axis`, and an `update_state()` refresh in `read()` so the
velocity is current rather than the last poll.

PR #288, opened 2026-08-24
feat(arm): passive leader and gripper force feedback

shantanuparab-tr:
Two opt-in arm options, both defaulting to today's behaviour. The gripper channel implements the reverse seam added in #285, so that contract now has a real implementer.

| Config | Effect |
| --- | --- |
| `actuated: false` | read-only arm: no staging, no teleop mode setup, no rest move |
| `gripper_force_feedback` + 3 curve constants | reflect the follower's grasp onto an actuated leader gripper, plain effort only |

**A passive leader's gripper goes to effort mode, not position.** Both source branches fall back to `Mode::position` whenever feedback is off, which makes a leader gripper hold its setpoint and fight the hand that is supposed to be driving it. It is an input and has to stay back-driveable. Zero effort is commanded explicitly, because a mode change alone leaves the previous setpoint and a stale non-zero one squeezes the gripper shut. Teardown releases it (zero, then idle), guarded by a flag since `end_teleop()` can be called with no preceding `prepare_for_teleop()` — the hardware-test park step does exactly that — and commanding effort on an idle gripper is a controller error.

**No joint frame remapping.** Selecting the `glide_left` / `glide_right` driver model already returns transformed values, so `joint_signs` / `joint_offsets`, `apply_joint_remap()` and `RemapKind` are all gone, and `trossen_arm_producer.{hpp,cpp}` is untouched by this PR.

**Conflict to settle before `experimental/trossen-webapp` merges.** That branch's lightweight leader renders gripper feedback through `set_gripper_external_effort` with no key to switch it; this PR renders through plain effort, matching what all five Rivet configs set and what the Glide handles were tuned on (external effort fights the operator on that hardware). The `gripper_feedback_mode` key is deliberately absent — its only consumer would be out of tree, the same reasoning that dropped four symbols under D7 — so that merge needs to re-add it deliberately rather than discover the difference on hardware.

**Scope:** both options are only settable through `configure()`, which needs a live controller, so nothing here is unit-tested; the curve was checked numerically at both tunings. Excludes `telemetry_frame_id()` (D1), `summon_joint()` (D13), and all haptic code and config (D4).

### DIFF
--- a/include/trossen_sdk/hw/arm/trossen_arm_component.hpp
+++ b/include/trossen_sdk/hw/arm/trossen_arm_component.hpp
@@ -7,6 +7,7 @@
 #define TROSSEN_SDK__HW__ARM__TROSSEN_ARM_COMPONENT_HPP_
 
 #include <memory>
+#include <optional>
 #include <string>
 #include <vector>
 
@@ -56,6 +57,7 @@ public:
    *   "staged_position": [0, 1.0, 0.5, 0.6, 0, 0, 0],  // optional, joint-space
    *   "staging_time_s": 2.0,       // optional, default 2.0 (stage / rest move)
    *   "write_moving_time_s": 0.1,  // optional, default 0.0 (per-tick smoothing)
+   *   "actuated": false,           // optional, default true (false = read-only arm)
    *
    *   // Host-side clamp on outgoing position commands. Optional; one entry per
    *   // joint, null leaves that joint unclamped. Applied before smoothing.
@@ -138,6 +140,15 @@ private:
   std::vector<float> read_joint();
   void               write_joint(const std::vector<float>& cmd);
 
+  /// Follower role: current measured gripper effort (N), or nullopt without a
+  /// driver. A sensor read, valid regardless of the gripper's control mode.
+  std::optional<float> read_gripper_effort();
+
+  /// Leader role: render gripper force feedback from the follower's measured
+  /// gripper effort (N) via the cubic curve. Only meaningful when
+  /// gripper_force_feedback_ is set.
+  void apply_gripper_feedback(float follower_gripper_effort);
+
   std::vector<float> read_cartesian();
   void               write_cartesian(const std::vector<float>& cmd);
 
@@ -156,6 +167,17 @@ private:
     }
     void write(const std::vector<float>& cmd) override {
       self->write_joint(cmd);
+    }
+    // Gripper force-feedback channel. Only the joint view carries it: the
+    // reflected force is a gripper effort, which has no cartesian analogue.
+    bool renders_gripper_feedback() const override {
+      return self->gripper_force_feedback_;
+    }
+    std::optional<float> read_gripper_effort() override {
+      return self->read_gripper_effort();
+    }
+    void apply_gripper_feedback(float follower_gripper_effort) override {
+      self->apply_gripper_feedback(follower_gripper_effort);
     }
   };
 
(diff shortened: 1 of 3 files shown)


shantanuparab-tr:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/288?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#288** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/288?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/288?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

include/trossen_sdk/hw/arm/trossen_arm_component.hpp
```diff
@@ -157,6 +168,17 @@ class TrossenArmComponent : public HardwareComponent,
...
+    // reflected force is a gripper effort, which has no cartesian analogue.
+    bool renders_gripper_feedback() const override {
+      return self->gripper_force_feedback_;
+    }
+    std::optional<float> read_gripper_effort() override {
+      return self->read_gripper_effort();
+    }
+    void apply_gripper_feedback(float follower_gripper_effort) override {
+      self->apply_gripper_feedback(follower_gripper_effort);
+    }
```
schromya, line 181: This should also be done in CartView. Not sure what the comment refers to as gripper force-feedback does have a cartesian analogue.

src/hw/arm/trossen_arm_component.cpp
```diff
@@ -334,6 +385,28 @@ void TrossenArmComponent::prepare_for_teleop() {
   // adaptive cutoff wide, and effectively disables the smoothing for the first
   // ticks of the session — when the arm is nearest the operator.
   cmd_filt_.reset();
+  if (!actuated_) {
+    // Passive leader: the arm joints have no motors, so putting them in
+    // position mode is harmless and keeps their positions readable. What we
```
schromya, line 390: "Keeping position readable" isn't really true. The reason we need to set the joints to a non-idle mode (either position or effort) is that the light on  the base is green for idle and blue for operating (position, effort, etc). If some of the joints are idle and the gripper is operating, then this light will flicker between blue and green which is very unpleasant to look at.

src/hw/arm/trossen_arm_component.cpp
```diff
@@ -334,6 +385,28 @@ void TrossenArmComponent::prepare_for_teleop() {
...
+    // tick instead of going slack and then stiffening.
+    driver_->set_gripper_mode(trossen_arm::Mode::effort);
+    driver_->set_gripper_effort(
+      gripper_force_feedback_ ? gripper_feedback_offset_ : 0.0, 0.0, false);
+    gripper_effort_engaged_ = true;
+    return;
+  }
   if (is_leader_) {
     // Leader: enable gravity compensation.
     driver_->set_all_modes(trossen_arm::Mode::external_effort);
```
schromya, line 412: If the leader is actuated and still uses force feedback, then when gripper effort is set in apply_gripper_feedback with `set_gripper_effort` (instead of `set_all_external_efforts`), an error will be thrown by the driver. This should be fixed here.

schromya (changes requested):
Changes requested.

PR #289, opened 2026-08-24
glide: add GlideSession input arbitration for Glide handles

shantanuparab-tr:
A Glide handle carries an analog joystick and four momentary buttons alongside its arm joints, and several components read the same physical handle for unrelated purposes — joystick to the mobile base, buttons to the session. Nothing stopped two of them binding the same input and each acting on it without knowing about the other.

`GlideSession` is the arbiter. Components declare what they consume, and an overlap throws at `configure()` instead of surfacing as two conflicting reactions to one button press at 30 Hz in the field. `GlideClaimLease` is the RAII holder, so a component cannot leak a reservation and block its own replacement after a reconfigure.

**No `libtrossen_arm` in the header.** Input arrives through a registered `std::function` reader, because the handle input API exists only on driver builds with lightweight-leader support. So this layer compiles anywhere, and PR 10 supplies the reader that knows better. Readers are invoked outside the lock — one handle blocking on hardware cannot stall another component's claims.

**Claims are advisory, not an access gate.** `read_inputs()` never consults the claim table; an unclaimed joystick is simply unread. Three consequences worth stating, none visible from the class:

1. `glide_arm_input` claims nothing — it only registers the reader. Consumers claim, never the producer.
2. `claims_` is touched only by `claim_inputs()` / `claim_holder()` / `release_claims()`.
3. Everything is keyed by `arm_id`, so one handle is a one-entry map.

Together those are why a single Glide with no base works unchanged, and why the process-global singleton costs the solo case nothing. Its only real limitation is two independent Glide *sessions* in one process, which no configuration wants.

- **Handle outputs are not here.** LED effects, brightness, and the vibration motor — `GlideOutputCommand`, `GlideOutputWriter`, `register_writer`, `set_button_led`, `set_led_brightness`, `set_vibration` — are all deferred as nice-to-have. Consequence upstack: PR 10 drops its `TROSSEN_HAVE_ARM_INPUT_COMMAND` probe (nothing left to gate) and PR 11 ships without button-LED feedback.
- **No test hooks.** `set_test_snapshot()` was public API on a process-global singleton whose value took precedence over live hardware — anything in-process could replace a real handle's input with a fake one, unreported. `register_reader()` already accepts any callable, so a test injects synthetic input the same way production does.
- **`GlideInput::kTrigger` dropped.** Its comment claimed anything else claiming it would collide loudly, but `TrossenArmComponent` never calls `claim_inputs()`, so no collision could ever be detected. It arrives with an actual claimant or not at all.
- **`GlideInputSnapshot::frame_id` dropped** — no remaining producer or consumer.
- **Button bits validated against the four physical buttons**, not the 32-bit mask they arrive in. A claim on bit 7 is now a configure error rather than a reservation that reads `false` forever.

**Verification**

Build clean, 398/398 ctest. Behaviour checked outside the suite: claim conflict and idempotence, per-handle scoping, atomicity on both rejection paths, component-wide release across handles, lease move/guard semantics, reader replacement and unknown-handle reads. Then ThreadSanitizer over 8 threads × 300 iterations, including a reader that reads another handle re-entrantly — no races.

**No tests in this PR.** The Glide suite is deferred with the rest of the extraction's tests, so this layer reaches a rig on hardware verification alone.

Also documents the handle's physical button layout on `kGlideButtonCount`: **0 = top, 1 = left, 2 = bottom, 3 = right**. Bits 1 and 3 were documented backwards until 2026-08-07, and a config written against the old note binds left and right swapped. Since the `.rst` was dropped from this stack, this header is the only place that layout is recorded.

### DIFF
--- a/include/trossen_sdk/hw/glide/glide_session.hpp
+++ b/include/trossen_sdk/hw/glide/glide_session.hpp
@@ -1,0 +1,273 @@
+/**
+ * @file glide_session.hpp
+ * @brief Shared input arbitration for Trossen Glide leader handles.
+ */
+
+#ifndef TROSSEN_SDK__HW__GLIDE__GLIDE_SESSION_HPP_
+#define TROSSEN_SDK__HW__GLIDE__GLIDE_SESSION_HPP_
+
+#include <cstdint>
+#include <functional>
+#include <initializer_list>
+#include <mutex>
+#include <optional>
+#include <stdexcept>
+#include <string>
+#include <string_view>
+#include <unordered_map>
+#include <utility>
+
+namespace trossen::hw::glide {
+
+/**
+ * @brief A logical input on a Glide handle.
+ *
+ * A Glide is a passive leader arm whose handle also carries an analog
+ * joystick and a set of momentary buttons. Several hardware components read
+ * the same physical handle for unrelated purposes — the arm joints drive a
+ * follower arm, the joystick drives the mobile base, the buttons drive the
+ * session — so each declares what it consumes and GlideSession rejects
+ * overlaps at configure() time rather than letting two components silently
+ * fight over one input.
+ */
+enum class GlideInput {
+  /// The 2-axis analog joystick, claimed as a unit (x and y are never split
+  /// between components — nothing sensible reads one axis in isolation).
+  kJoystick,
+
+  /// One momentary button, identified by its bit position in the handle's
+  /// button bitmask. Claims on this input must carry a `bit`.
+  kButton,
+};
+
+std::string_view glide_input_name(GlideInput input);
+
+/**
+ * @brief One input a component wants to reserve on a handle.
+ *
+ * For `kButton`, `bit` selects which button — the index of its bit in the
+ * handle bitmask. For every other input `bit` is unused and left at -1.
+ */
+struct GlideClaim {
+  GlideInput input;
+  int        bit{-1};
+};
+
+/// Convenience: a claim on button `bit`.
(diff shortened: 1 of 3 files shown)


shantanuparab-tr:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/289?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#289** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/289?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/289?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

include/trossen_sdk/hw/glide/glide_session.hpp
```diff
@@ -0,0 +1,273 @@
+/**
+ * @file glide_session.hpp
```
schromya, line 2: I really dislike this. It is really stretching the architecture too much. But I can't think of a better way to do it without severely changing the entire architecture.

schromya: As per our discussion, I'll pause this review until we have discussed if this is the best way to go about it.

schromya (commented):
Pausing review of this stack per our discussion.

PR #290, opened 2026-08-24
glide: publish handle joystick and buttons from the arm driver

shantanuparab-tr:
`GlideArmInputComponent` fills the reader seam added in #289. For each handle it names, it resolves the already-constructed `TrossenArmComponent`, borrows its driver, and registers a `GlideInputReader` that turns the driver's input report into an SDK-owned snapshot. It is the only place that knows handle input arrives over the arm driver — every mapping component downstream reads snapshots and never touches one.

**The reader captures the arm by** **`weak_ptr`, and that is load-bearing.** `GlideSession` is process-global and outlives any one session, so a strong capture would pin the arm and its open controller connection past shutdown. Arm controllers are single-client, so the leak would stall the _next_ run's connect until the controller timed the stale client out. The destructor unregistering every reader is the other half of the same guarantee.

**Read failures are swallowed, logged once per handle.** The reader runs on the teleop mirror and session-control poll threads, where an escaping throw would take teleop down over one dropped input frame. Consumers read `nullopt` as "this handle contributed nothing this tick", which for the base means its axes read zero.

**It claims nothing.** Claims are made by the components that _consume_ an input — never by the one that publishes it — so a handle whose joystick nothing reads is simply unread. That is what lets a solo Glide with no base work unchanged.

**New** **`hardware.controls`** **section**, keyed by component id, declaring this component and PRs 11–12's:

```json
"controls": {
  "glide_inputs": { "type": "glide_arm_input", "arms": ["glide_left", "glide_right"] }
}
```

A control binds a physical control surface to a software effect. The checkable property separating the section from `arms` and `cameras` is that **none of its members is recorded and none appears in** **`producers`**. Three shape decisions, each a deliberate departure from a neighbouring section:

| Decision | Neighbour does | Why different |
| --- | --- | --- |
| keyed map | `cameras` is an array | `merge_overrides()` is object-only, so `--set` cannot reach into an array (`type_error.305`). Verified: `--set hardware.controls.session_control.poll_rate_hz=80.0` works |
| `std::map` | `arms` is `unordered_map` | Deterministic construction order. Ordering here is already implicit and unenforced, so it should not also depend on a hash seed |
| only `type` modelled | `CameraConfig` models width/height/fps | The three control types share no other settings; modelling any would mean a member per key of one type, unread by the other two. Uses the parse-and-erase pattern `camera_config.hpp`'s own TODO recommends rather than its `known_keys` list |

[...]

PR #291, opened 2026-08-24
glide: drive a mobile base from Glide handle sticks and buttons

shantanuparab-tr:
`GlideBaseComponent` is a base-space teleop leader — the `BaseSpaceTeleop` interface from #284, now with a leader as well as a follower. `read()` samples the configured handle inputs into `[linear, angular, lift, lateral]`; `write()` is a no-op.

Separate from the arms because the handles drive a different follower than the arm joints do. Modelling them as one bundled device would force a single teleop pair to carry both, so each role is its own pair over the same physical handles, and #289's arbiter keeps them from reading the same input.

**The mapping is entirely declarative, and that is a hardware fact rather than a style preference.** The driver reports each stick only as "0 to 4095" with no physical direction, and the button bit assignments are documented nowhere the SDK can see. Every axis assignment and invert flag was arrived at by driving the base and watching it. So discovering that "raise" is really bit 4 has to be a config edit, not a rebuild. The class docblock records the trap: **the two handles are mirrored models and do not agree**, so neither can be inferred from the other — assuming otherwise produced two wrong signs before the flags converged. It also keeps the bring-up note that if translation direction changes between bring-ups *with nothing edited*, the cause is the swerve modules' homed zero landing half a turn out, and a sign flip will only be right until the next restart.

**One read per handle per tick, not one per axis.** Per-axis reads could mix a left-handle sample from one driver cycle with a right-handle sample from the next, which on a swerve base shows up as rotation and translation disagreeing about when the operator moved. Verified: three axes across two handles produces exactly two reads.

**Buttons are a three-state axis** — full speed either way, or stopped. Both held cancels rather than picking one, so a stuck button cannot run the lift into its end stop.

**Two config mistakes rejected that the claim table cannot catch.** #289 makes repeated identical claims from one component idempotent by design, so the lease is blind to a component overlapping itself:

| Rejected | Why it would otherwise be silent |
| --- | --- |
| `up_bit == down_bit` | `sample_axis()` cancels up against down, so the axis reads zero however hard it is pressed |
| two axes binding one button | one press drives both, and the second claim is accepted as a duplicate |

Deliberately buttons-only. Several axes sharing one handle's **joystick** is the intended pattern — the stick is claimed as a unit, so `linear`-on-x plus `lateral`-on-y is two axes over one claim.

- **Holonomic `translation` is not here** — it lands in the next PR in the stack. A differential-drive base needs only the independent axes in this one, which is what makes this a complete unit. Split because the two together came to 680 lines.
- **No input-freshness gating.** `input_timeout_ms`, the per-handle frame-id record, and the staleness drop in `read()` are all deferred. **Consequence worth knowing before this runs on a rig:** on a Rivet the handles are reached over Wi-Fi sharing a radio with the video feed, and the measured failure is not a clean disconnect — the UDP telemetry loses 60–70% in bursts while the link stays associated and the driver keeps answering *with the last frame it received*. So a read succeeds in microseconds and hands back a joystick position from seconds ago, and the base drives on it. `TeleopController::Config::leader_timeout_ms` does not catch this: the reads keep returning perfectly on time. Change-detection on the axes is not a substitute either — a stick held steady and a stick nobody can see produce identical numbers.
- **No button LEDs.** The `led_*` config keys, `apply_led_effects()` and `update_led_brightness()` are gone with the handle output path, so the lift buttons stay dark instead of lighting while their axis is live.
- **`deadzone` is now validated on independent axes too**, not only on `translation`. Previously `{"max": 1.0, "deadzone": 5.0}` was accepted and then silently zeroed every command — the exact failure the translation check existed to catch. One shared helper now covers both, so a bad number reads the same wherever it appears.
- **`get_info()` reports the resolved axis map** — arm id, source, invert, max, deadzone, and the button bits. The original reported a subset alongside LED state. Given the plan records that the shipped example config disagrees with both rig presets on all three axes, being able to print what actually got parsed is worth having at bring-up.

**Verification.** Build clean, 398/398 ctest. Checked outside the suite with a synthetic reader: centre reads zero, full deflection reads exactly `max`, `invert` flips sign, the output-units deadzone (0.5 of a 2.0 max ⇒ 0.20 travel gives 0 and 0.30 moves), buttons three-state including both-held-cancels and unmapped bits ignored, a handle with no reader contributing zero rather than faulting, claims recorded per handle and released on destruction, one read per handle per tick, and all nine `configure()` rejection paths.

**No tests in this PR.** The Glide suite is deferred with the rest of the extraction's tests. **Coverage note:** neither Workbench config declares a `glide_base` — only the Rivet example does, and that is the one that cannot be built here. So this compiles ungated but no configuration we can run exercises it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,7 @@ add_library(trossen_sdk SHARED
   src/hw/base/slate_base_component.cpp
   src/hw/base/slate_base_producer.cpp
   src/hw/glide/glide_arm_input_component.cpp
+  src/hw/glide/glide_base_component.cpp
   src/hw/glide/glide_session.cpp
   src/hw/teleop/teleop_controller.cpp
   src/hw/teleop/teleop_factory.cpp
--- a/include/trossen_sdk/hw/glide/glide_base_component.hpp
+++ b/include/trossen_sdk/hw/glide/glide_base_component.hpp
@@ -1,0 +1,185 @@
+/**
+ * @file glide_base_component.hpp
+ * @brief Mobile-base teleop leader driven by Glide handle joysticks and buttons.
+ */
+
+#ifndef TROSSEN_SDK__HW__GLIDE__GLIDE_BASE_COMPONENT_HPP_
+#define TROSSEN_SDK__HW__GLIDE__GLIDE_BASE_COMPONENT_HPP_
+
+#include <array>
+#include <optional>
+#include <string>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#include "nlohmann/json.hpp"
+
+#include "trossen_sdk/hw/glide/glide_session.hpp"
+#include "trossen_sdk/hw/hardware_component.hpp"
+#include "trossen_sdk/hw/teleop/teleop_capable.hpp"
+
+namespace trossen::hw::glide {
+
+/**
+ * @brief Base-space teleop leader that reads Glide handle inputs.
+ *
+ * A separate component from the arms on purpose. The operator's hands are on
+ * two Glide handles that each carry a joystick and buttons alongside the arm
+ * itself, and those inputs drive a different follower (the mobile base) than
+ * the arm joints do (the follower arms). Modelling them as one bundled device
+ * would force a single teleop pair to carry both, so instead each role is its
+ * own pair over the same physical handles, and `GlideSession` guarantees no two
+ * components read the same input.
+ *
+ * Leader-only: `read()` samples the configured inputs, `write()` is a no-op.
+ *
+ * ### Axis mapping is configuration, not code
+ *
+ * Which stick axis drives which base axis, and which button bits raise and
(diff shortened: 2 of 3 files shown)


shantanuparab-tr:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/291?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#291** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/291?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/291?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #292, opened 2026-08-24
glide: treat one stick as a 2D translation vector for holonomic bases

shantanuparab-tr:
Second half of the Glide base leader, on top of #291. That PR gave every base axis its own independent mapping; this one adds the mode a holonomic base actually wants.

**Why independent axes are wrong for a swerve base.** A stick driving two axes is one 2D input, and treating it as two scalars breaks in two measurable ways:

| | Independent axes | `translation` |
| --- | --- | --- |
| dead region | a **square** — a diagonal nudge inside the corner reads zero while the same magnitude on-axis moves | a **circle** — same threshold in every direction |
| full diagonal | `max * sqrt(2)`, ~41% faster than straight ahead | exactly `max` |

Measured on the shipped tuning (`max: 0.6`, `deadzone: 0.05`):

```
straight = 0.6      diagonal = 0.6      (independent axes would give 0.848528)
on-axis |0.20| = 0.0763638     diagonal |0.20| = 0.0763365
just outside the deadzone -> 0.00443582 m/s
```

Row 2 is the radial deadzone: a diagonal and an on-axis push of equal magnitude produce the same speed to four decimal places. Row 3 is the rescale — the surviving range is mapped back onto 0..1, so output is continuous from zero at the deadzone edge. Without it the base would lurch to 8% of max the instant the stick left the dead region. Direction is preserved by scaling the unit vector, so 45° in gives 45° out through the clamp.

**Mutual exclusion, not precedence.** `translation` owns `linear` and `lateral`, so setting either in `axes` alongside it is rejected rather than letting one silently win — the config author is told to remove one so it is unambiguous which produces the command. `angular` and `lift` are untouched and pair with `translation` normally.

A differential-drive base has no lateral axis to pair, so it keeps using the independent entries from #291 and nothing about that form changes.

**Verification.** Build clean, 398/398 ctest, and #291's two harnesses still green against this build. Additionally checked: centre reads zero on both axes, straight-forward reaches `max`, the eight `translation`-specific rejection paths (missing `arm_id`, a `buttons` source, an unknown source name, both sources on the same stick axis, non-positive `max`, `deadzone >= max`, and the collision with each of `linear` and `lateral`), that `translation` claims the stick as a unit while a second component claiming the same stick is refused by #289's arbiter, and that `translation` alone is a valid config with no `axes` block at all.

[...]

PR #293, opened 2026-08-24
glide: drive session control from Glide handle buttons

shantanuparab-tr:
**This is the implementer #286 has been missing.** `SessionControlCapable` shipped as a seam once `KeyboardComponent` was dropped as a nice-to-have; `GlideSessionControlComponent` fills it. A poller thread watches the claimed handle buttons and emits a semantic intent on the rising edge, so an operator with both hands on the handles can start, re-record or end a session without letting go of the robot — which is what lets the cockpit display stay a read-only heads-up view.

Its stop button emits the same `kStopSession` the webapp's Stop button does, because the same physical button drives both hosts.

**Edge semantics.** Rising edge only: a held button is one intent rather than one per poll, and release is not an event of its own. A change must persist through the debounce window before it counts, so contact bounce on a momentary switch cannot start an episode twice.

**Bit assignments are configuration** for the same reason as #291's axis map — the handle's button bitmask layout is documented nowhere the SDK can see — so re-binding a button is a config edit, not a rebuild. The header cross-references `kGlideButtonCount` for the physical cross the four bits form (0 top, 1 left, 2 bottom, 3 right), since a config author binding bit 1 should know which button they just picked.

**Two ordering requirements, both now stated rather than assumed**

- `set_callbacks()` must precede `start()`. The poller reads the stored callback without synchronisation, so installing one on a running poller is a data race. That is how `attach_session_control()` drives it, but nothing enforced or said it.
- The destructor joins the poller **before** the claim lease releases, so the thread cannot read an input the component no longer holds. That works only because `lease_` is declared after the thread members and destruction runs in reverse. The note is on the member itself, not just the class docblock — a tidy-up reorders members, not comments.

- **`emit_event()` does not exist.** The extraction called a base-class helper; #286's contract is `set_callbacks(on_event, on_disconnect)` with the implementation storing the callbacks. So the component gains an `EventCallback` member and the override.
- **`summon` is gone** from `event_from_name()` / `event_name()` and from the documented event list — it is not in the shipped enum.
- **`poll_once()` is private.** It was public purely as a test seam ("inject a snapshot, call this, assert what was emitted"), and #289 removed `GlideSession::set_test_snapshot()`, so nothing can drive it. Its docblock no longer advertises a workflow that does not exist.
- **`on_disconnect` is accepted and deliberately never called.** This source cannot distinguish "permanently unusable" from "quiet right now": a failed handle read yields `nullopt` and recovers on the next frame, so reporting a disconnect would end the session over a dropped packet. `(void)` with the reason in a comment, rather than a stored member nothing reads.
- **Duplicate bindings rejected** — new. The claim table cannot see it, because repeated identical claims from one component are idempotent by design. Worse here than in #291, since the two events can contradict: `button 0 on handle 'glide_left' is bound to both 'start' and 'stop_session'; one press would emit both`.
- **`configure()` commits all state at the end** — new. The original assigned `poll_rate_hz_` and `debounce_` to the members before parsing the button list, so a config rejected on a bad event name left the component with a new poll rate and its old buttons.

**Verification.** Build clean, 398/398 ctest, and **ThreadSanitizer clean** with the poller live — the first PR in this stack to ship a thread of its own. Behaviour checked against the built library:

| Checked | Result |
| --- | --- |
| held button | emits once, not per poll |
| release | not an event |
| second press | emits again |
| per-bit mapping / unbound bits | correct / ignored |
| two bits rising together | both emit |
| debounce 80 ms | nothing at 30 ms, one event by 150 ms |
| 10 ms transient | swallowed |
| `start()` / `stop()` twice | idempotent; no events after stop |
| destructor | poller joined, then claim released |
| absent handle | silent, no fault |
| no callback installed | survives |

Plus all nine `configure()` rejection paths, including a bit outside the four physical buttons, which #289's claim validation catches.

**No tests in this PR** — the Glide suite is deferred with the rest of the extraction's tests, so the checks above are not in ctest. This completes the Glide input layer: the arbiter (#289), the publisher (#290), the base leader (#291/#292), and now the session-control consumer.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,7 @@ add_library(trossen_sdk SHARED
   src/hw/glide/glide_arm_input_component.cpp
   src/hw/glide/glide_base_component.cpp
   src/hw/glide/glide_session.cpp
+  src/hw/glide/glide_session_control_component.cpp
   src/hw/teleop/teleop_controller.cpp
   src/hw/teleop/teleop_factory.cpp
   src/hw/camera/opencv_camera_component.cpp
--- a/include/trossen_sdk/hw/glide/glide_session_control_component.hpp
+++ b/include/trossen_sdk/hw/glide/glide_session_control_component.hpp
@@ -1,0 +1,150 @@
+/**
+ * @file glide_session_control_component.hpp
+ * @brief Session control driven by Glide handle buttons.
+ */
+
+#ifndef TROSSEN_SDK__HW__GLIDE__GLIDE_SESSION_CONTROL_COMPONENT_HPP_
+#define TROSSEN_SDK__HW__GLIDE__GLIDE_SESSION_CONTROL_COMPONENT_HPP_
+
+#include <atomic>
+#include <chrono>
+#include <string>
+#include <thread>
+#include <vector>
+
+#include "nlohmann/json.hpp"
+
+#include "trossen_sdk/hw/glide/glide_session.hpp"
+#include "trossen_sdk/hw/hardware_component.hpp"
+#include "trossen_sdk/hw/session_control/session_control_capable.hpp"
+
+namespace trossen::hw::glide {
+
+/**
+ * @brief Turns Glide handle button presses into session-control intents.
+ *
+ * The operator teleoperating a Rivet has both hands on the Glide handles, so
+ * reaching for a keyboard to start or re-record an episode means letting go of
+ * the robot mid-task. This component maps handle buttons onto the same
+ * `SessionControlEvent` intents any other source emits, which is what lets the
+ * cockpit display be a read-only heads-up view rather than something the
+ * operator has to drive.
+ *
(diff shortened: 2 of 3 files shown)


shantanuparab-tr:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/293?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#293** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/293?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/293?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #295, opened 2026-08-24
examples: add trossen_workbench, bimanual Glide teleop

shantanuparab-tr:
**The first end-to-end exercise of the Glide input layer that can be built here.** The Workbench is the Rivet layout without the mobile base — two Glide handles driving two follower arms, bolted to a fixed bench. Nothing touches `trossen_base`, so it builds in the default configuration with no private dependency, unlike the Rivet example further up the stack.

Three files, matching the house style every other example follows: the main, a config, and a README. No `BRINGUP.md`, no second config — `main` has neither anywhere.

It exercises, on one binary: the arbiter (#289), the input publisher (#290), the base leader's absence, the button mapper (#293), the arm command shaping from #287, and the passive-leader path from #288.

**The handles are passive leaders.** No actuators, moved by hand, joints never commanded. Their grippers are the exception — effort mode, so they stay back-driveable while rendering the follower's grasp force to the operator's fingers.

**Buttons drive the session**, bound in config rather than code because the button bitmask layout is documented nowhere the SDK can check:

| Button | Bit | Event |
|---|---|---|
| Top | 0 | `start` |
| Left | 1 | `stop_session` |
| Bottom | 2 | `rerecord` |

**Deviations from the source branch**

- **`hardware.controls`, not `hardware.components`** — the keyed section from #290 rather than the generic bucket that was dropped. Verified `--set hardware.controls.session_control.debounce_ms=80` reaches it, which is what the map shape bought.
- **`attach_control(*sc); sc->start();` → `attach_session_control(sc);`** — the double start goes, and #286's `shared_ptr` signature lands. `attach_session_control()` calls `set_callbacks()` then `start()` in that order, which is exactly the ordering #293's header requires, so the example gets it right by construction.
- **Added a guard for more than one session-control source.** The manager rejects a second by throwing, and an exception escaping `main()` terminates before `mgr` unwinds, leaving producer threads unjoined. Prints and returns 1 instead, matching the observer-failure pattern already in the file.
- **`joint_signs` / `joint_offsets` and `gripper_feedback_mode` dropped from the config.** The first pair went with the driver-side remap; `gripper_feedback_mode` has no consumer in tree. Worth stating because `ArmConfig` has **no `extra` passthrough**: leaving any of them in would be silently discarded while looking meaningful. A check asserts the config carries none of the five keys this stack removed.
[...]

PR #294, opened 2026-08-24
arm: accept every model and end effector the driver knows

shantanuparab-tr:
`TrossenArmComponent::configure()` accepted exactly **one** model — `wxai_v0` — and **two** end effectors. The installed driver has **seven** models and **nineteen** named end effectors, so a config naming a Pro follower or a Glide handle was rejected before the driver ever saw it:

```
TrossenArmComponent: Unknown model: glide_right
```

Every example on `main` uses `wxai_v0`, which is why nothing caught it. Found by running the Workbench example (next PR in the stack) — its config is the first in the repo to name `pro`, `glide_left` or `glide_right`, and **all three** were rejected. This is not Glide-specific: a `pro` arm could not be configured either.

**Models resolve against the driver's own map.** `trossen_arm::MODEL_NAME` is a `Model -> string` map the driver publishes, so looking the name up there means the accepted set cannot drift when the driver adds a model. That is the property that matters here — a hardcoded list is precisely how these two cases went missing.

**End effectors are tabled by hand, and the comment says why.** The driver has no `END_EFFECTOR_NAME` equivalent: they are `static constexpr` members with no names attached. So this list has to be extended manually when the driver adds a gripper — the opposite of the model path, and worth stating rather than leaving a reader to wonder why the two differ.

The table covers both naming styles the driver offers: the five undated aliases (`wxai_v0_leader`, `no_gripper`, …), which resolve to the driver's current revision and are what a config should normally name, and the fourteen revision-pinned names (`wxai_v0_leader_20260626`, `pro_base`, …) for a rig that must not move when an alias does.

**`is_leader` is a column, not an inference.** It selects gravity compensation (`Mode::external_effort`) over position mode in `prepare_for_teleop()`, so it is behaviour rather than a label. Only the `_leader` variants are hand-guided; a `_base` or `_follower` gripper is commanded. A name-suffix heuristic would have been shorter and would have quietly mis-set it the first time a gripper broke the pattern.

**Both rejections now list the valid set**, which turns a typo into a fix instead of a search through driver headers:

```
Unknown model: nonsense (valid: wxai_v0, vxai_v0_right, vxai_v0_left, core, pro,
glide_left, glide_right)
```

[...]

PR #296, opened 2026-08-24
base: Rivet swerve drivetrain as a base-space teleop follower

shantanuparab-tr:
## Summary

`TrossenBaseComponent` — the Rivet's mobile base as a `teleop::BaseSpaceTeleop` follower. Holonomic swerve drive plus a vertical lift, honouring all four `base_axis` axes where the differential-drive SLATE honours two.

Plus the `TROSSEN_ENABLE_RIVET` option and the `hardware.mobile_base` schema entry that declares this hardware.

## Why there is a servicing thread, and why `command_timeout_ms` exists

This is the part that isn't reconstructable from the code, so it belongs here.

The driver must be ticked whether or not teleop is running — a leader that stops writing must not stall the base mid-motion. So the component owns a thread ticking `update_base()` at a fixed rate, independent of teleop.

That thread is exactly why the timeout is needed. The wheels hold their last commanded velocity and the thread re-asserts it every tick, so **a base told to drive keeps driving until something says otherwise**, whether or not anything is still commanding it. A mirror loop that dies, a session that ends without `end_teleop()`, a host process wedged on something else: all of them leave a moving robot with nobody at the controls, and none of them are visible from here.

So the servicing thread also checks how old the last accepted command is and commands a standstill past `command_timeout_ms` (default 500 ms, against a mirror that writes at 1 kHz — three orders of magnitude of headroom). It arms on the **first** `write()`: before that there is nothing to go stale, and a caller that configures the base without ever driving it shouldn't be told its link is dead.

**It does not cover a handle whose values go stale while still flowing.** That check is deferred; `read()` would keep returning the last stick value, `write()` would keep arriving with fresh timestamps, and this watchdog would never fire. The two are not substitutes.

## Four different things are called `trossen_base`

The name is kept for compatibility with shipped rig configs, so the header carries the disambiguation:

| `trossen_base` | is |
| --- | --- |
| `#include "trossen_base/trossen_base.hpp"` | the external drivetrain library |
| the CMake target | that same library, as a dependency |
| this class | an SDK *consumer* of that library |
| `"type": "trossen_base"` in config | the selector for this component, and the MCAP stream key |

It is also **not** the SLATE base — that one is differential drive with no lift, and a config naming the wrong one connects to hardware that is not there.

## `TROSSEN_ENABLE_RIVET` gates the library, not the robot

The option means "is the `trossen_base` library available to link against", which is why only this component sits behind it. The Glide handle layer needs `libtrossen_arm` and nothing else, so **a Workbench builds and runs the whole handle path with the option off**.

`trossen_base` is consumed with `find_package(trossen_base REQUIRED)` beside `libtrossen_arm` — both Trossen-shipped prebuilts, found rather than built — so a host without it fails at configure time naming the package rather than at link time naming a symbol.

## Schema

`hardware.mobile_base` gains a `type` field defaulting to `"slate_base"`, the way `CameraConfig::type` defaults to `"realsense_camera"`, plus an `extra` passthrough. The base-specific keys ride in `extra` and reach `configure()` verbatim, so each base keeps its defaults in one place — its own `configure()` — rather than having them restated in the config struct where they would drift apart.

## Test Plan

- [x] Default configuration (`TROSSEN_ENABLE_RIVET=OFF`): builds clean, **398/398 tests pass**, zero `TrossenBaseComponent` symbols in the library
- [x] `TROSSEN_ENABLE_RIVET=ON` against an **installed** `trossen_base`: configures, builds, links — 114 symbols present, `ldd` resolves `libtrossen_base.so` from the install prefix
- [x] `TROSSEN_ENABLE_RIVET=ON` with `trossen_base` **absent**: fails at configure with `Could not find a package configuration file provided by "trossen_base"` — no hang, no link-time surprise
- [x] cpplint / pre-commit clean

## Notes for review

- The e-stop **behaviour** is intact: `write()` commands a standstill while `driver_->is_e_stopped()`, so releasing the e-stop can't hand the wheels a velocity from before it was pressed. What this PR omits is the e-stop/telemetry/battery *read-out* API, which had no in-tree consumer.
- Pre-existing bug found and deliberately **not** fixed here: `MobileBaseConfig::enable_torque` defaults to `false` while `SlateBaseComponent::configure()` defaults it to `true`, and `to_json()` emits it unconditionally — so a SLATE config that omits the key gets torque off through the struct and on through raw JSON. It predates this stack and only affects SLATE.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,12 @@ option(TROSSEN_ENABLE_RERUN_OBSERVER "Build RerunObserver (depends on rerun-cpp 
 # server bridge flow: cmake -DTROSSEN_SDK_ENABLE_POLICY_CLIENT=ON ...
 option(TROSSEN_SDK_ENABLE_POLICY_CLIENT
   "Build PolicyClient (depends on IXWebSocket + msgpack-cxx)" OFF)
+# OFF by default: it requires the trossen_base drivetrain library to be installed,
+# which most hosts do not have. It gates that LIBRARY, not the robot -- the Glide
+# handle layer needs only libtrossen_arm and is built unconditionally, so a
+# Workbench runs the whole handle path with this off. Enable on a Rivet:
+#   cmake -DTROSSEN_ENABLE_RIVET=ON ...
+option(TROSSEN_ENABLE_RIVET "Enable Rivet mobile base support (requires trossen_base)" OFF)
 
 add_library(trossen_sdk SHARED
   src/backends/backend_registry.cpp
@@ -130,6 +136,17 @@ find_package(BZip2 REQUIRED)
 find_package(Arrow REQUIRED)
 find_package(Parquet REQUIRED)
 find_package(libtrossen_arm REQUIRED)
+
+# Rivet's drivetrain library, consumed the same way as libtrossen_arm above: a
+# Trossen-shipped prebuilt, found rather than built. REQUIRED so a host without
+# it fails at configure time naming the package, instead of at link time naming
+# a symbol.
+if(TROSSEN_ENABLE_RIVET)
+  message(STATUS "Rivet mobile base support enabled")
+  find_package(trossen_base REQUIRED)
+  target_sources(trossen_sdk PRIVATE src/hw/base/trossen_base_component.cpp)
+  target_link_libraries(trossen_sdk PUBLIC trossen_base)
+endif()
 
 include(FetchContent)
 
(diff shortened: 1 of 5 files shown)


shantanuparab-tr:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/296?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#296** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/296?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/296?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #297, opened 2026-08-25
base: record Rivet base odometry and commanded velocity

shantanuparab-tr:
## Summary

`TrossenBaseProducer` — records the Rivet base's odometry and the velocity it was commanded, plus the recording surface that needs: `lift_velocity` on the existing `Odometry2DRecord`, the matching `Odometry2D.proto` field, and the backend write hunk.

Split from PR 14 because recording base telemetry is independent of the teleop path.

## The twist channel is an action echo, not an observation

The base reports **no measured velocity**. Pose comes from its own odometry, but the recorded twist is the velocity most recently *commanded* to it, read back from the component (which is the only thing that knows it).

For a learned policy that distinction is the difference between an observation and an action, so it is stated in the header, at the record, and in the stream metadata rather than left for a dataset consumer to infer.

`base_velocity_names` advertises all four axes the swerve base drives — `linear_vel, angular_vel, lift_vel, lateral_vel` — so a converter does not assume the SLATE's two-axis differential-drive layout.

## `lift_velocity` sits in `Odometry2D`, and that is provisional

A lift is not planar motion, so it goes outside the 2D twist. But a record named `Odometry2D` carrying it at all is a compromise: the same argument would drag in every other non-planar thing a base measures.

A `TODO` on the record says so and names the two ways out — rename to something that admits it describes base state generally, or move the non-planar axes to their own record.

## Battery telemetry is deliberately not recorded — and that removed a data race

It does not belong in a record called `Odometry2D`. Investigating it turned up a second reason.

`TrossenBase` caches everything into **one unsynchronised `base_data_` struct**:

- `get_pose()` writes its *pose* members
- `update_base()` → `refresh_battery_status()` writes its *battery* members

Disjoint members are distinct memory locations, so the producer's `poll()` and the component's 15 Hz servicing thread do **not** race, and the reads underneath both take the driver's own `status_mutex_` against its receive thread.

**But `get_percent()`, `get_temp()` and `get_charging_state()` read exactly the members that servicing thread writes**, with nothing synchronising them. Calling one from `poll()` is a genuine data race on a non-atomic float, in the recording path. Not recording battery is what keeps this poll confined to disjoint state.

The analysis is recorded at `poll()`, framed as a constraint on what may be added there — the next person's instinct will be "the driver exposes battery, why aren't we recording it."

### This answers an open question from the plan

The known SLATE hazard is that `SlateBaseProducer::poll()` drives `update_state()` itself, racing the component's `set_cmd_vel()` on the driver's serial transactions. **This pair is not shaped that way** — the producer never ticks the driver; the component's servicing thread does, unconditionally. The producer is a pure reader, and safe, but only because of which fields it touches.

## Test Plan

- [x] Default configuration: **399/399**, including a new test asserting `lift_velocity` reaches the `.mcap` — both that the embedded schema declares the field and that the serialised payload carries the value
- [x] **Negative control run**: that test fails with the write hunk removed, passes with it restored
- [x] `TROSSEN_ENABLE_RIVET=ON` against an installed `trossen_base`: builds and links, 125 producer symbols
- [x] cpplint / pre-commit clean

The test matters because a SLATE never sets `lift_velocity`. A mistake in either half of its path is invisible on every existing rig and would surface only as a silently zeroed column in a Rivet dataset, long after recording.

## Notes for review

- `schema_data_odom2d_` is built at runtime from the compiled descriptor pool, not a checked-in blob, so the embedded schema cannot drift from the `.proto`. Verified rather than assumed.
- Pre-existing quirk hit while writing the test: the backend maps only the **empty string** to `McapCompression::None`; `"none"` falls through to the "Unknown compression option" warning and then uses None anyway. The test uses `""` with a comment rather than emitting a spurious warning. Not fixed here.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,7 +144,10 @@ find_package(libtrossen_arm REQUIRED)
 if(TROSSEN_ENABLE_RIVET)
   message(STATUS "Rivet mobile base support enabled")
   find_package(trossen_base REQUIRED)
-  target_sources(trossen_sdk PRIVATE src/hw/base/trossen_base_component.cpp)
+  target_sources(trossen_sdk PRIVATE
+    src/hw/base/trossen_base_component.cpp
+    src/hw/base/trossen_base_producer.cpp
+  )
   target_link_libraries(trossen_sdk PUBLIC trossen_base)
 endif()
 
--- a/include/trossen_sdk/data/record.hpp
+++ b/include/trossen_sdk/data/record.hpp
@@ -112,6 +112,15 @@ struct JointStateRecord : public RecordBase {
 
 /**
  * @brief 2D odometry state (pose + velocity), mirroring nav_msgs/Odometry.
+ *
+ * TODO(shantanuparab-tr): Revisit this record's name and shape. `lift_velocity`
+ * below is not 2D odometry and does not mirror nav_msgs/Odometry either — it is
+ * vertical motion carried here because a base that reports it has nowhere else
+ * to put it, and the same argument would drag in every other non-planar thing a
+ * base measures. Either this becomes a record whose name admits it describes
+ * base state generally, or the non-planar axes move to a record of their own.
+ * Battery telemetry was proposed alongside the lift and deliberately left out
+ * for exactly this reason.
  */
 struct Odometry2DRecord : public RecordBase {
   /// @brief 2D pose in the odom frame.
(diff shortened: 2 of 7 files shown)


shantanuparab-tr:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/297?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#297** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/297?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/297?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #298, opened 2026-08-25
examples: trossen_rivet, the Workbench plus a mobile base

shantanuparab-tr:
## Summary

`trossen_rivet` — the Workbench example plus a mobile base. Two Glide handles driving two follower arms **and** a holonomic swerve base with a vertical lift.

Three files, matching the house style every other example follows: `trossen_rivet.cpp` + `config.json` + `README.md`.

## Each handle drives two followers at once

The part worth understanding before reading the code:

| The handle's… | drives | via |
|---|---|---|
| joint motion | the follower arm on that side | a joint-space teleop pair |
| joystick and buttons | the mobile base | a base-space teleop pair |

`GlideSession` arbitrates the inputs so no two components read the same stick or button — a config that double-binds one is rejected at startup rather than producing two things moving from one press.

## Derived from the shipped Workbench, not from older Rivet sources

The two mains share ~90%, so this is the Workbench main plus four base-specific additions. Porting an older Rivet main instead would have reintroduced four things this stack already fixed: `hardware.components`, the old `attach_control` signature, the missing multi-source guard, and the removed `generate_episode_path` helper.

What the base adds:

1. **Construction from `hardware.mobile_base`** through the registry by `type` — so the same code brings up a `slate_base`. This is the consumer for the `type` field added in #296.
2. **A producer branch** for the base.
3. **A teleop pair** (`base_leader` → `trossen_base`, space `base`).
4. **A teardown that halts the wheels.**

**Ordering, deliberately:** the base is constructed **last**. `configure()` waits for the base to report ready and then re-homes the swerve modules — tens of seconds, and the slowest step in the whole bring-up. Everything else has already failed fast by then, so a config typo does not cost a homing cycle before it is reported.

**Teardown matters more here than for the arms.** The wheels *hold* their last commanded velocity and the servicing thread re-asserts it every tick, so a base left mid-command keeps driving. Reached through `BaseSpaceTeleop` rather than the concrete type, so it stays correct for any base the config names.

A missing `mobile_base` is not fatal — the arms and handles alone are a working Workbench — but it says so out loud, because a Rivet whose base silently did not appear looks like a base that failed to move.

## config.json is regenerated, not copied

Built programmatically from the shipped Workbench config plus the base, so the two examples cannot drift on what they share.

**Stripped, all verified absent:** `haptic_*` (D4), `input_timeout_ms` (D8), `led_*` (D15), `estop_battery_percent` (D22), and `leader_timeout_ms` — whose stall watchdog is deferred (D1), so the preset's key would configure nothing.

**Command clamps ported in** — the example had none, which looked like an oversight rather than a decision. Using the **exact-2° values PR 07 ships** (`1.5358897 / 3.1066861 / 2.3212879`), not the preset's, which differ in the 4th decimal. **J0 is hand-measured and mirrored** — left `−1.1…0.8`, right `−0.8…1.1`. It is the shoulder-yaw stop keeping the arms off each other and the chassis, and the one bound that must not be copied across; mirroring it wrong points both arms into the same space.

**The base's component id is its `type`.** `hardware.mobile_base` is singular with no `id` field, so the id has to come from somewhere; this follows the existing SLATE precedent exactly, where `trossen_mobile_ai` uses id == `hardware_id` == `stream_id` == `"slate_base"`.

## The README is new, and corrects two things the old docs had wrong

This example has never had a `README.md` — its documentation was `BRINGUP.md` plus `GLIDE_CONTROLS.md`, neither of which exists in this tree. Checking every operator-facing claim against the generated config found two errors worth not shipping:

- `GLIDE_CONTROLS.md` says *"The other buttons on either handle do nothing on this robot."* **False** — the left handle's bits 0/1/2 are start / stop\_session / rerecord. An operator following it would not know how to start an episode.
- Its closing section documents the **lift buttons being lit**, dim-then-bright. Those LEDs are gone with D15.

The button map is stated as the physical cross with the **0 = top, 1 = left, 2 = bottom, 3 = right** warning, including that bits 1 and 3 were documented backwards until 2026-08-07 and that nothing validates a bit beyond its range.

Also folded in, because these are the questions this rig actually generates: why the diagonal is not faster than straight ahead (radial dead zone, not two independent axes); why the base halts itself, **and that it does not catch a handle that goes quiet while still reporting its last stick value**; that a direction change between bring-ups with nothing edited is the homed zero landing half a turn out, so do not chase it with sign flips; and that the base's twist is a command echo rather than a measurement.

## Test Plan

- [x] `TROSSEN_ENABLE_RIVET=ON` (installed `trossen_base`): clean full build from scratch, binary produced, loads its config, `--dump-config` renders it
- [x] `--set hardware.mobile_base.max_linear_mps=0.3` reaches the new keys (the object-keyed schema from D19; an array here would throw `type_error.305`)
- [x] Default configuration: **399/399**, and **zero** rivet targets — the gate excludes the example, not just the library
- [x] `"space": "base"` verified against `kSpaceDescriptors`; clamp arrays verified length 7 with J0 mirrored
- [x] pre-commit clean

## Notes for review

- Resolves the `TODO: RENAME?? @schromya` from the old example header by **deciding**: the name stays `trossen_rivet`, matching the product, the rig hostnames (`rivet-01`, `rivet-02`) and the preset it derives from.
- The IP table is `rivet-01`'s; `rivet-02` uses a different assignment, and the README says so rather than presenting one rig's wiring as universal.
- **Inconsistency worth a decision:** `trossen_workbench` (#295) drives the same two `pro` followers with **no command clamps**. The plan scoped the clamp port to this PR, but arm-on-arm collision applies to a Workbench too — only the chassis half of the rationale is Rivet-specific. Adding the same clamps there is a small follow-up if wanted.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -17,6 +17,14 @@ target_link_libraries(trossen_mobile_ai trossen_sdk libtrossen_arm)
 add_executable(trossen_workbench trossen_workbench/trossen_workbench.cpp)
 target_link_libraries(trossen_workbench trossen_sdk libtrossen_arm)
 
+# Trossen Rivet — the Workbench plus a holonomic swerve base with a lift.
+# Gated because the base needs the trossen_base library; the Glide handle layer
+# it shares with the Workbench is not gated and builds either way.
+if(TROSSEN_ENABLE_RIVET)
+  add_executable(trossen_rivet trossen_rivet/trossen_rivet.cpp)
+  target_link_libraries(trossen_rivet trossen_sdk libtrossen_arm)
+endif()
+
 # Trossen Stationary AI Policy — 2 followers + 4 cameras, policy-driven leaders.
 # Requires the PolicyClient gate (FetchContent of IXWebSocket + msgpack-cxx).
 #
--- a/examples/trossen_rivet/README.md
+++ b/examples/trossen_rivet/README.md
@@ -179,23 +179,25 @@ chase it by flipping signs; a flip will only be right until the next restart.
 The handles and the followers are on **different subnets** — deliberate, and it
 matches the rig wiring, so check both before assuming a fault.
 
-| Arm | Role | Default IP |
-|---|---|---|
-| `glide_left` | Handle (passive leader) | `192.168.0.3` |
-| `glide_right` | Handle (passive leader) | `192.168.0.2` |
-| `follower_left` | Follower | `192.168.1.4` |
-| `follower_right` | Follower | `192.168.1.5` |
+| Arm | Role | Default IP | Interface |
+|---|---|---|---|
+| `glide_left` | Handle (passive leader) | `192.168.5.13` | wireless |
+| `glide_right` | Handle (passive leader) | `192.168.5.12` | wireless |
+| `follower_left` | Follower | `192.168.1.15` | wired |
+| `follower_right` | Follower | `192.168.1.14` | wired |
 
-These match `rivet-01`. **Other rigs differ** — `rivet-02` uses a different
-assignment — so confirm against the rig rather than assuming, and override
-without editing the file:
+These match `rivet-02`, the rig this example was brought up on. **Other rigs
+differ** — `rivet-01` uses a different assignment — so confirm against the rig
+rather than assuming, and override without editing the file:
 
 ```bash
 ./build/examples/trossen_rivet \
-  --set hardware.arms.glide_left.ip_address=192.168.5.13
+  --set hardware.arms.glide_left.ip_address=192.168.5.3
 ```
 
 > Arms do not answer ICMP, so `ping` is not a valid reachability test. Use `arp`.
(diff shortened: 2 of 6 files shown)


shantanuparab-tr:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/298?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#298** <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/298?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/TrossenRobotics/trossen_sdk/298?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

